Issue #182, opened 2026-05-31
feat(web): rich hover card on a running issue's codex/activity cell


corylanou:
Part of #138. Hovering a running issue's codex/activity cell (the CODEX UPDATE / EVENT column) should show a **rich hover card** — today the text is CSS-truncated with no hover, so the latest codex activity is unreadable.

## What
- Hover (and keyboard focus) on the codex/activity cell of a running row → a popover/card showing: **full latest message** (un-truncated), **recent activity** (last ~3-5 events with timestamps), and **current turn / tokens (in/out) / runtime** for that session.
- Lightweight popover (CSS / tiny JS, HTMX-SSE friendly), themed + dark mode, positioned so it never clips at the viewport edge.
- Accessible: focusable, aria-described, dismiss on blur/esc.

## Acceptance
- Hovering/focusing a running row's codex cell shows the full latest message + recent activity + turn/tokens/runtime; no clipping; works light + dark + narrow widths.

## Files
- internal/web/templates (running rows + hover card), small popover JS; data from telemetry.Running (#135).

corylanou:
## Codex Workpad

Status: Done

Summary:
- PR #187 merged the rich running-activity hover/focus card for #182.
- After #187 merged, the automated review thread on recent activity messages was addressed in follow-up PR #190.
- PR #190 is merged and records event-specific recent activity for token usage/rate-limit/process/turn events instead of reusing stale agent text.

Acceptance:
- Running Codex/activity cells expose a hover and keyboard-focus card with the full latest message, recent timestamped activity, turn count, token in/out totals, and runtime.
- The card uses fixed viewport positioning, Escape/blur dismissal, light/dark styling, and was checked at desktop and narrow mobile widths.

Validation:
- Focused tests: `go test ./internal/web/templates ./internal/telemetry ./internal/orchestrator ./internal/runner ./internal/web`.
- Review-fix focused test: `go test ./internal/runner`.
- Full gate: `make check` passed before #187, and passed again for #190 after rebasing onto `origin/main` with 80.1% coverage.
- Browser preview: desktop dark and mobile light hover cards stayed within the viewport; Escape closed the card.
- CI: #187 checks passed; #190 checks passed and the PR merged.

Review threads:
- Original #187 automated review thread was replied to and resolved after #190 merged.
- #190 has no review threads.

Blockers: none.

PR #188, opened 2026-05-31
fix(web): stabilize help popovers

### LINKED ISSUE #185: fix(web): help tooltips flicker badly + rewrite every hint to explain the *why* (not just a definition)
Follow-up to #173. Two problems with the '?' help tooltips:

## 1. Flicker (interaction bug)
Hovering a '?' makes the hint card **flicker in and out and jump around even with the mouse held still**. Classic causes — fix all of them:
- The popover appears **at/over the cursor** or **captures pointer events**, so showing it fires mouseleave on the trigger → hide → mouse back on trigger → show → infinite flicker loop. Fix: `pointer-events: none` on the card (or only interactive when intended), and position it **anchored to the '?' element, offset so it never sits under the cursor**.
- It **repositions on every mousemove** → jumping. Fix: compute placement once on open; don't follow the cursor.
- No **hover-intent / bridge**: add a small show delay + keep-open grace so moving from the '?' toward the card doesn't dismiss it; close on blur/esc/outside-click.
- Make it a **single shared, stable, accessible tooltip/popover utility** used by ALL '?' hints — and reuse the same utility for the codex hover card (#182) so neither flickers and they look consistent.
- Viewport-aware placement (flip/shift so it never clips at edges).

## 2. Hint copy must explain the WHY, not just define the term
Every hint is currently technically accurate but meaningless to someone who doesn't know the internals. Rewrite **all** of them as: **what it is + why it matters + when to care**, in plain language. Examples of the bar to hit:
- **Codex rate limits** → 'How much of your Codex API quota is left before requests get throttled. When it runs low, agents slow down or pause — watch it during heavy parallel work.' (not 'the API rate limits')
- **Throughput (tps)** → 'Tokens processed per second across all running agents — a live pulse of how hard Symphony is working right now.'
- **Budget** → 'Optional spend caps. If set, Symphony refuses to start new work once a daily or per-issue dollar limit is hit — your guardrail against runaway cost.'
- **Backoff / Retry queue** → 'Work that failed and is waiting to be retried after a cooldown, so a transient error doesn't drop the task.'
- **Blocked** → 'Issues Symphony can't proceed on without a human (missing dependency, auth, or a needed action).'
- **Runtime** → 'How long this Symphony instance has been running.'
- Do the same for Running, Queue, Completed, Tokens (in/out), turn/age, session, Board health, Agent activity, Primary/Secondary/Credits buckets, etc.
Keep them to ~1-2 short sentences; lead with the meaning, not the mechanism.

## Acceptance
- Hovering any '?' shows a **stable** card (no flicker/jump, held still and reachable); accessible (keyboard + aria); never clips at edges.
- Every hint reads as a plain-language 'what + why + when to care'; a non-expert understands what they're looking at and why it's shown.

## Files
- internal/web/templates (shared tooltip utility + the hint content map), small popover JS/CSS; align with #182.

corylanou:
## Summary

- Replaced per-help hover handling with a shared delegated popover utility for help cards and issue hover cards.
- Kept cards pointer-transparent by default, anchored to their trigger, viewport-aware, keyboard accessible, and closed by Escape/outside click/blur.
- Rewrote help copy to explain what each metric means, why it matters, and when to care.

Fixes #185

## Test Plan

- [x] `go test ./internal/web/...`
- [x] Browser check on `http://localhost:<ephemeral>/reports` for stable desktop and mobile tooltip placement
- [x] `make check`


corylanou:
Conflicts with main — #182's hover card landed in `internal/web/templates/dashboard.templ` since this branched. Please: rebase onto `origin/main`, resolve conflicts in `dashboard.templ` and `help.*` keeping BOTH the hover-card and the new help-popover/hint-copy changes, run `make generate` then `make check`, and force-push.

PR #200, opened 2026-05-31
fix(web): fingerprint static assets

### LINKED ISSUE #198: fix(web): static assets have no cache-busting/headers — browsers serve stale CSS/JS after a deploy
**Found while dogfooding** — caused real confusion: after merging a dashboard fix + rebuilding/relaunching, a maintainer hard-refreshed and still saw the OLD behavior (flickering help), because the browser served a stale stylesheet.

## Problem
Static assets are served at stable URLs with **no content hash and no cache headers**:
- HTML references `<link rel="stylesheet" href="/static/css/output.css">` (and similarly JS) — no `?v=` / no fingerprint.
- Responses for `/` and `/static/css/output.css` send **no `Cache-Control` / `ETag` / `Last-Modified`** (verified via `curl -D -`), so browsers apply heuristic caching and may keep the old CSS/JS across deploys. A normal — and often even a hard — refresh won't reliably bust it.

Result: a user can be on the new binary but rendered with old CSS/JS → "the fix isn't there."

## Fix (pick one, ideally both)
- **Fingerprint static asset URLs**: emit `/static/css/output.<hash>.css` (content hash) and reference that hash from templates, so a new build = a new URL = guaranteed fresh fetch. (Build-time hash or an in-memory hash computed at startup.)
- **Send correct cache headers**: fingerprinted assets → `Cache-Control: public, max-age=31536000, immutable`; non-fingerprinted HTML → `Cache-Control: no-cache` (or `ETag` + revalidate). At minimum add `ETag`/`Last-Modified` so conditional requests revalidate.

## Acceptance
- After a rebuild/relaunch, a normal browser refresh shows the new CSS/JS without manual cache-clearing.
- Static assets carry a content fingerprint and/or proper cache headers; HTML is revalidated.

## Files
- internal/web (static file handler + cache headers), internal/web/templates (asset URL references), build/generate step if hashing at build time.

corylanou:
## Summary
- Compute startup content fingerprints for static assets and serve hashed `/static/...` aliases with immutable cache headers.
- Render dashboard, settings, reports, and onboarding with fingerprinted stylesheet URLs.
- Mark HTML and legacy static URLs with revalidation headers so refreshes pick up new deploys.

Fixes #198

## Test Plan
- [x] `go test ./internal/web -run 'TestServerStaticAssetsUseFingerprintsAndCacheHeaders|TestServerLegacyStaticAssetsRequireRevalidation'`
- [x] `go test ./internal/web/...`
- [x] `make check`

### DIFF
--- a/internal/web/assets.go
+++ b/internal/web/assets.go
@@ -1,0 +1,182 @@
+package web
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+	"io/fs"
+	"net/http"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/labstack/echo/v4"
+
+	symphony "github.com/digitaldrywood/symphony"
+	"github.com/digitaldrywood/symphony/internal/web/templates"
+)
+
+const (
+	staticURLPrefix        = "/static/"
+	defaultStylesheetPath  = "/static/css/output.css"
+	headerETag             = "ETag"
+	fingerprintHashLength  = 12
+	immutableCacheControl  = "public, max-age=31536000, immutable"
+	revalidateCacheControl = "no-cache"
+)
+
+type staticAssets struct {
+	fsys          fs.FS
+	originals     map[string]staticAsset
+	fingerprinted map[string]staticAsset
+}
+
+type staticAsset struct {
+	urlPath       string
+	filePath      string
+	fingerprinted string
+	etag          string
+}
+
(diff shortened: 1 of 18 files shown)


corylanou:
Conflicts after #201 (help-tooltip rework) merged — both touch dashboard.templ + input.css/output.css. Please rebase onto origin/main, keep BOTH the cache-busting asset changes AND #201's help-tooltip host/markup, run make generate + make check, force-push.

PR #217, opened 2026-06-01
feat(config): hot-reload global config

### LINKED ISSUE #211: feat(config): hot-reload global.yaml — live add/remove projects without restart
## Problem
Per-project `WORKFLOW.md` is already hot-reloaded (#41) via `internal/config/watcher`, but the top-level **`global.yaml`** is read **once at boot** (`internal/cli/boot.go`). Adding or removing a project in the `projects:` list — the common "I want to point Detent at another repo" case — currently requires restarting the orchestrator.

## What
- Watch the active `--config` `global.yaml` path (reuse the `internal/config/watcher` fsnotify + debounce pattern from #41).
- On change: reload + validate `globalconfig`, diff the `projects:` list against the running `project.Manager`, and reconcile:
  - new project -> `Manager.Add` (`internal/project/manager.go:134`)
  - removed project -> `Manager.Remove` (`internal/project/manager.go:145`)
  - changed project settings (weight/priority/workdir) -> apply (remove+add or a targeted update)
- Reload global knobs live where safe (e.g. `max_concurrent_agents`, `scheduling`); for any field that genuinely needs a restart, log a clear warning instead of silently ignoring it.
- On an invalid edit: log the validation error and **keep the current running config** (never crash the orchestrator mid-flight).

## Acceptance
- With `detent` running, append a new project block to `global.yaml` -> the project appears on the dashboard and begins dispatching, **no restart**.
- Remove a project block -> that project stops cleanly; in-flight agents for it are handled per existing Remove semantics.
- A malformed `global.yaml` save is logged and ignored; the previously-good config stays live.
- Table-driven tests for the reconcile diff (added/removed/changed/unchanged, invalid-config retention).

## Files
- `internal/cli/boot.go` (wire a global-config watcher alongside the per-project ones)
- `internal/config/watcher` (generalize or add a global variant)
- `internal/project/manager.go` (Add/Remove already exist)

corylanou:
## Summary

- watch the active global.yaml and reconcile running projects on valid edits
- add manager diff/reconcile support for added, removed, changed, and unchanged projects
- keep the last good config live when reload or reconcile fails

Fixes #211

## Test Plan

- [x] `make check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,9 +376,10 @@ detent --config ~/.detent/global.yaml promote <id> --priority 1
 detent --config ~/.detent/global.yaml remove-project <id>
 ```
 
-These commands persist the global config. Restart a foreground Detent process
-after structural edits unless your deployment wires the live manager signal path
-into the same process.
+These commands persist the global config. A running Detent process watches the
+active `global.yaml` and reconciles project additions, removals, and setting
+changes without a restart. Invalid edits are logged and ignored while the last
+valid config stays live.
 
 ## Dashboard And APIs
 
--- a/internal/cli/boot.go
+++ b/internal/cli/boot.go
@@ -167,6 +167,11 @@ func startRunning(ctx context.Context, cfg BootConfig) error {
 	if err := manager.Start(runCtx); err != nil {
 		return err
 	}
+	globalWatcherDone := startGlobalConfigWatcher(runCtx, cfg.Global, manager, logger)
+	defer func() {
+		stop()
+		waitGlobalConfigWatcher(globalWatcherDone)
+	}()
 
 	snapshotHub := hub.New[telemetry.Snapshot]()
 	go publishSnapshots(runCtx, manager.Registry(), snapshotHub, runtimeStore, displayURL, defaultSnapshotInterval, time.Now)
--- a/internal/cli/boot_test.go
+++ b/internal/cli/boot_test.go
@@ -149,8 +149,6 @@ func TestRegistryRefresherRequestsProjectOrchestrators(t *testing.T) {
 }
 
 func TestStartRunningBootsDashboardAndStopsOnContextCancel(t *testing.T) {
-	t.Parallel()
-
 	host, port := freeLoopbackPort(t)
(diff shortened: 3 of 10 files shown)


corylanou:
@codex review

corylanou:
Moving to **Rework** — the re-review surfaced **two P1-orange** correctness bugs in the hot-reload path (`internal/project/manager.go`):
1. **Keep old project until replacement starts** — if the replacement project is constructed but `Start` fails inside `registerProjectLocked`, the existing project has already been stopped+deleted, so a bad reload edit takes the project offline (contradicts "ignore invalid edits, keep last good config live").
2. **Keep reconcile from partially applying failed reloads** — a failed reconcile must be all-or-nothing; don't leave the manager in a half-applied state.

Make reload atomic: validate + construct + Start the replacement **before** tearing down the old project, and on any failure roll back to the previously-live project. Add table tests for: replacement Start fails, mid-reconcile failure, invalid config retained.

corylanou:
@codex review

corylanou:
Back to **Rework** — the fresh review found a third ordering bug (now the replacement starts *before* the old project stops, so both run concurrently and can double-dispatch). The swap needs an exact order that avoids **both** failure modes at once. Use this sequence in `replaceStartedProjectLocked` / the changed-project path:

1. **Construct** the replacement project (validate config, build connector/orchestrator) **without starting it**.
2. If construction/validation fails → keep the old project running, reject the edit, retain last-good config. (no offline)
3. **Stop** the old project and wait for it to quiesce.
4. **Start** the replacement. (never concurrent — old is fully stopped first)
5. If Start fails → roll back: attempt to restart the old project, surface the error, keep last-good config.

Apply the same construct-before-mutate, all-or-nothing discipline to the multi-project reconcile (removes+adds): validate/construct everything that can fail up front, then commit the changes; on any failure roll back so the registry always matches the last-good config. Add table tests for: construct fails (old stays live), Start fails after stop (rollback restarts old), and a reconcile where a later add fails (earlier removes rolled back). Run make check, then I'll re-gate on a fresh Codex pass.

corylanou:
@codex review

PR #239, opened 2026-06-01
fix(errors): preserve joined error chains

### LINKED ISSUE #231: fix: preserve error chains with %w in joined errors
## Problem
A few sites join two errors but wrap only the first with `%w`, formatting the second with `%v` — dropping it from the chain so `errors.Is/As` cannot match:
- `internal/codex/transport.go:125` — `fmt.Errorf("%w: close stdin: %v", ctx.Err(), closeErr)`
- `internal/codex/transport.go:168` — `fmt.Errorf("%w: kill process: %v", ctx.Err(), killErr)`
- `internal/cli/doctor.go:653` — `...close sqlite database: %v", err, closeErr)`

`transport.go:153` already uses `%w` correctly — this is an inconsistency. Go 1.20+ supports multiple `%w` in one `fmt.Errorf`.

## Fix
Use `%w` for both errors so both are inspectable via `errors.Is/As`.

## Acceptance
- Test asserts `errors.Is` matches the secondary (close/kill) error on the wrapped result.
- `make check` green.

## Files
internal/codex/transport.go, internal/cli/doctor.go

corylanou:
## Summary

- Wrap secondary close/kill errors with `%w` so joined error chains remain inspectable.
- Add focused regression coverage for transport and SQLite doctor cleanup error chains.

Fixes #231

## Test Plan

- [x] `go test ./internal/codex`
- [x] `go test ./internal/cli`
- [x] `make check`

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -649,12 +649,16 @@ func openDoctorSQLite(ctx context.Context, path string) (doctorStore, error) {
 	db.SetMaxOpenConns(1)
 
 	if err := db.PingContext(ctx); err != nil {
-		if closeErr := db.Close(); closeErr != nil {
-			return nil, fmt.Errorf("ping sqlite database: %w; close sqlite database: %v", err, closeErr)
-		}
-		return nil, fmt.Errorf("ping sqlite database: %w", err)
+		return nil, doctorSQLitePingError(err, db.Close())
 	}
 	return db, nil
+}
+
+func doctorSQLitePingError(err, closeErr error) error {
+	if closeErr != nil {
+		return fmt.Errorf("ping sqlite database: %w; close sqlite database: %w", err, closeErr)
+	}
+	return fmt.Errorf("ping sqlite database: %w", err)
 }
 
 func runDoctorCommand(ctx context.Context, path string, args ...string) error {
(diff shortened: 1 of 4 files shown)


corylanou:
**Parked (Blocked) — Windows-only test failure agents can't reproduce.** `Verify (windows-latest)` fails (GoReleaser Snapshot is just the downstream cascade):

```
--- FAIL: TestLocalTransportCloseWrapsKillErrorAfterContextCancellation
transport_test.go:230: Close() error = context canceled: kill process: invalid argument, want process done error in chain
```

On Windows, `Process.Kill()` on an already-exited process returns `invalid argument` rather than the 'process done' error the new test asserts via the chain. The `%w` production change in `Close()` is fine; the **new test is platform-specific**. Fix: make the assertion tolerant — assert the killErr is wrapped regardless of value (e.g. check the message contains `kill process:`, or capture the actual killErr and `errors.Is` against it), or guard the strict chain check with `runtime.GOOS != "windows"`. Codex agents don't run Windows CI so they can't converge on this — needs a human/Windows-aware fix.

corylanou:
Moving to **Rework** to fix this in-system. The `%w` production change is correct; only the new test is platform-specific. Make `TestLocalTransportCloseWrapsKillErrorAfterContextCancellation` Windows-tolerant per the diagnosis above — assert the kill error is wrapped regardless of its value (check the message contains `kill process:`, or capture the actual `killErr` and `errors.Is` against it), or guard the strict chain assertion with `if runtime.GOOS != "windows"`. Then rebase onto `origin/main`, run `make check`, and re-request review.

PR #248, opened 2026-06-02
fix(orchestrator): redispatch active issues

### LINKED ISSUE #245: fix(orchestrator): on startup, re-dispatch active-state issues that have no live session
## Problem
When Detent restarts while agents are mid-flight (e.g. an operator rebuild/relaunch), the running agent subprocesses are killed. Their issues remain in active states (`In Progress`/`Rework`/`Merging`) on the board, but the new orchestrator process does **not** re-dispatch them — it only claims fresh `Todo` issues. The work is stranded: the issue sits `In Progress` with an open PR and no running agent, and `POST /api/v1/refresh` does not recover it.

Observed tonight: a restart while 5 idiom-fix agents were running orphaned all 5 (issues #230/#231/#232/#233/#234 stuck In Progress, no session, never re-dispatched). They had to be hand-gated/merged.

## Fix
On startup (and on reconcile), for each issue in an active state (`In Progress`, `Rework`, `Merging`) that has **no live session/PID** in the orchestrator, re-dispatch it as a continuation (resume from the existing worktree + PR per the WORKFLOW 'For In Progress' flow). Apply a small backoff to avoid thundering-herd on restart.

## Acceptance
- Restart the orchestrator with N in-flight agents; after restart, the N In-Progress issues are re-dispatched (continuation) rather than stranded.
- Table/integration test: state with In-Progress issues + no sessions ⇒ dispatch selects them.

## Files
- internal/orchestrator (dispatch/reconcile/startup)

corylanou:
## Summary

- Allow `In Progress` continuation issues with open PRs to dispatch when no live orchestrator session exists
- Stagger continuation redispatches during restart/reconcile to avoid launching every recovered agent at once
- Add dispatch and startup regression coverage for active-state redispatch

Fixes #245

## Test Plan

- [x] `go test ./internal/orchestrator -count=1`
- [x] `make check`

### DIFF
--- a/internal/orchestrator/dispatch_test.go
+++ b/internal/orchestrator/dispatch_test.go
@@ -416,7 +416,11 @@ func TestDispatchReadyIssuesStaggersContinuationDispatches(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	orch.dispatchReadyIssues(ctx, &state, []connector.Issue{first, second}, now)
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		orch.dispatchReadyIssues(ctx, &state, []connector.Issue{first, second}, now)
+	}()
 
 	request := receiveWorkerHostRunRequest(t, runner.started)
 	if request.Issue.ID != first.ID {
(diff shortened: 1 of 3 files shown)


corylanou:
@codex review

PR #299, opened 2026-06-02
fix(github): default blank project statuses asynchronously

### LINKED ISSUE #295: fix(github): projectItemStatusOrDefault blocks first orchestrator tick for minutes when many items have blank status
## Summary

PR #268 introduced `projectItemStatusOrDefault`, which calls `setProjectItemStatus` (a GraphQL mutation) **sequentially for every blank-status item** inside `fetchProjectItems`. With many unset items, this blocks the entire first orchestrator tick until all mutations complete — and in TTY (terminal dashboard) mode, the process dies with `context deadline exceeded` before the tick finishes.

## Reproduction

1. Have a GitHub ProjectV2 board with ≥10 issues that have no `Status` value set
2. Start `detent` in a TTY (e.g. a tmux session) so the Bubble Tea dashboard launches
3. Detent dies with `context deadline exceeded` after exactly 10 minutes — no issues are updated

In non-TTY mode (`detent 2>&1 | tee log`), the process survives but the first tick takes ~6 minutes (19 blank items × ~18–20 s each), during which the board state is `null` and no log output is emitted.

## Root Cause

`fetchProjectItems` calls `normalizeProjectItem` for every project item. The new `projectItemStatusOrDefault` path calls `setProjectItemStatus` (a live GraphQL write) for each item missing a status — sequentially, inline, blocking the scan:

```go
// internal/connector/github/adapter.go
func (c *Connector) projectItemStatusOrDefault(ctx context.Context, item projectItemNode) (string, *time.Time, error) {
    statusName := singleSelectName(item.StatusValue)
    if statusName != "" {
        return statusName, singleSelectUpdatedAt(item.StatusValue), nil
    }
    // ↓ synchronous GraphQL mutation per blank item — blocks the entire tick
    if err := c.setProjectItemStatus(ctx, itemID, statusName); err != nil {
        return "", nil, fmt.Errorf("default github status: %w", err)
    }
    return statusName, nil, nil
}
```

With 19 blank items and ~18–20 s per mutation call (empirically observed; individual mutations succeed in ~0.6 s when called directly), the scan takes ~6 minutes.

In TTY mode, a 10-minute deadline somewhere in the startup path (likely the Bubble Tea context) fires before the tick completes, and the process exits with `context deadline exceeded`.

## Evidence

- `detent.log` shows nothing between `"starting web server"` (T+0) and `"stopping web server"` (T+10 min) — orchestrator emits zero log lines during the blocked tick
- Goroutine dump (SIGQUIT on non-TTY run, T+13 min) shows orchestrator goroutine at `orchestrator.go:160` (the main `select` loop), confirming the first tick had just completed after ~6 minutes
- `board.state_distribution` in `/api/v1/state` remained `null` throughout both TTY and non-TTY runs
- After manually setting all 19 blank items to Backlog via the GitHub API, the board populated normally on the next poll

## Suggested Fix

Move the status-defaulting out of the per-item scan path and into a dedicated pass that:
1. Collects all item IDs that have blank status
2. Writes the Backlog default in parallel (or batched) after the scan completes
3. Does not block `FetchCandidateIssues` — either fire-and-forget with a goroutine, or run before the main scan as a separate pre-flight step

This keeps the scan fast and prevents one slow/failing mutation from breaking the entire tick.

## Environment

- `detent dev` (installed via `go install github.com/digitaldrywood/detent/cmd/detent@latest` on 2026-06-02, after PR #268 merged)
- macOS 25.5.0, arm64
- 20 open issues, 19 with blank `Status`
- `global.yaml` polling interval: 30 s

michaelhvisser:
## Summary
- treat blank GitHub ProjectV2 statuses as Backlog during project item scans without blocking on writes
- move blank-status default mutations into a bounded background worker with in-flight dedupe
- add a regression test that blocks the default write path and verifies scanning still returns promptly

Fixes #295

## Test Plan
- go test -run TestConnectorFetchIssuesByStatesDefaultsBlankProjectStatusesWithoutBlockingScan ./internal/connector/github
- go test ./internal/connector/github
- go build ./...
- go test ./...
- go vet ./...
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run --timeout=5m
- golangci-lint run --timeout=5m --new-from-rev=origin/main
- go test -race ./internal/connector/github

### DIFF
--- a/internal/connector/github/adapter.go
+++ b/internal/connector/github/adapter.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/digitaldrywood/detent/internal/connector"
@@ -19,6 +20,8 @@ const (
 	pullRequestsPageSize          = 100
 	pullRequestsPageLimit         = 3
 	defaultProjectItemStatusState = "Backlog"
+	defaultProjectItemStatusJobs  = 4
+	defaultProjectItemStatusTTL   = 2 * time.Minute
 )
 
 const projectItemsQuery = `
@@ -1305,6 +1308,7 @@ func attachMatchingPullRequests(
 func (c *Connector) fetchProjectItems(ctx context.Context, keepIssue func(connector.Issue) bool) ([]connector.Issue, error) {
 	var after *string
 	allIssues := []connector.Issue{}
+	statusDefaults := []projectItemStatusDefault{}
 
 	for {
 		var response struct {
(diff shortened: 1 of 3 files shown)


michaelhvisser:
Closing this draft as a duplicate of #300.

PR #316, opened 2026-06-05
feat(observability): show build metadata

### LINKED ISSUE #315: feat(observability): show build version + commit SHA + date in STATUS block and dashboard
## Goal

Show the **version, short commit SHA, and build date** the running binary was built from — in the TUI `DETENT STATUS` block and the web dashboard — so an operator can always see exactly what build is live (even for `dev` builds).

Today the data exists but isn't displayed, and local builds aren't stamped:
- `cmd/detent/main.go` already has `version`/`commit`/`date` ldflag vars, and the **Makefile already injects them** (`-X main.version=$(VERSION) -X main.commit=$(COMMIT) -X main.date=$(DATE)`, where VERSION=`git describe --tags --always`, COMMIT=`git rev-parse --short HEAD`).
- But `detent version` shows `version: dev / commit: none / build date: unknown` when built with a raw `go build` (no ldflags) — e.g. dev/air builds and ad-hoc rebuilds. And the STATUS block / dashboard don't surface any of it.

## Tasks

1. **Stamp every build path**, not just `make build`:
   - Ensure the **air/dev** target (`make dev`) and any documented `go build` use the same `LDFLAGS` so dev builds carry the SHA (e.g. `version=dev`, `commit=<short-sha>`, `date=<RFC3339>`). A raw `go build ./cmd/detent` with no ldflags should at least fall back to git via `debug.ReadBuildInfo()` `vcs.revision`/`vcs.time` (Go stamps these automatically with `-buildvcs`) so the SHA shows even without explicit ldflags.
   - Recommended: read `runtime/debug.ReadBuildInfo()` as a fallback when the ldflag vars are empty/`dev`/`none` — that gives commit + dirty flag + build time for free on plain `go build`.
2. **Surface it in the TUI `DETENT STATUS` block** — add a line near `Generated:`/`Runtime:`, e.g.:
   `Build: v0.3.0 (a1b2c3d) 2026-06-05T21:00Z` — or for dev: `Build: dev (a1b2c3d, dirty) 2026-06-05T21:00Z`. File: `internal/tui/model.go` (status render).
3. **Surface it on the web dashboard** — a small muted line in the navbar/footer or the health area (consistent with the slim navbar). File: `internal/web/...` status/health render.
4. Keep `detent version` output as-is (already correct when stamped).

## Acceptance
- A binary built via `make build`, `make dev`, and a plain `go build` all show a real commit SHA (release tag when present, `dev`+sha otherwise) — never `none`/`unknown` when built from a git checkout.
- The TUI `DETENT STATUS` block and the web dashboard each display version + short SHA + build date.
- `make check` green; a test asserting the status block includes the build/version line.

## Notes
- Operator action when this lands: rebuild with ldflags (`make build`, not raw `go build`) and restart so the running instance reports its real SHA.

corylanou:
## Summary

- add build metadata resolution from ldflags, Go build info, and a matching git checkout fallback
- show build version, short SHA, dirty marker, and date in the TUI status block and dashboard topbar
- route Air dev builds through the stamped Makefile build path

Fixes #315

## Test Plan

- [x] `go test ./internal/buildinfo ./internal/tui ./internal/web ./internal/web/templates ./cmd/detent ./internal/cli`
- [x] `go build -o tmp/detent-plain ./cmd/detent` then `./tmp/detent-plain version`
- [x] `make build VERSION=dev` then `./tmp/detent version`
- [x] `make check`

### DIFF
--- a/.air.toml
+++ b/.air.toml
@@ -6,15 +6,15 @@ tmp_dir = "tmp"
 
 [build]
   args_bin = []
-  bin = "./tmp/symphony"
-  cmd = "go build -o ./tmp/symphony ./cmd/symphony"
+  bin = "./tmp/detent"
+  cmd = "make build VERSION=${DETENT_AIR_VERSION:-dev}"
   delay = 1000
   exclude_dir = ["tmp", "vendor", "testdata", "node_modules", "internal/database/sqlc"]
   exclude_file = ["go.sum"]
   exclude_regex = ["_test\\.go", "_templ\\.go$"]
   exclude_unchanged = false
   follow_symlink = false
-  full_bin = "./tmp/symphony"
+  full_bin = "./tmp/detent"
   include_dir = []
   include_ext = ["go", "templ", "sql", "css", "js", "html", "yaml", "yml", "toml"]
   include_file = []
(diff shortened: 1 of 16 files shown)


corylanou:
@codex review

PR #319, opened 2026-06-06
feat(update): handle go-install updates

### LINKED ISSUE #317: feat(update): offer to run go-install OR swap to release binary instead of refusing (interactive + safe)
## Problem

`detent update --yes` **dead-ends** when the binary is managed by `go install`:

```
This Detent binary appears to be managed by go install. Run the Go install command instead.
Run: go install github.com/digitaldrywood/detent/cmd/detent@latest
Error: update refused
```

Refusing-by-default is the right instinct (don't silently clobber a `go install`-managed binary), but it leaves the user at a hard stop. The updater already *detected* the install source and *knows* the exact command — it should offer to act, not just print and exit non-zero.

## Desired behavior (interactive, with consent + warning)

When `install_source == go_install` and an update is available, **prompt** (don't refuse) with two actionable choices + abort:

1. **Run the Go install for me** — execute `go install github.com/digitaldrywood/detent/cmd/detent@latest` on the user's behalf, stream output, report the resulting version. (This is the safe, in-kind update — keeps it go-install-managed.)
2. **Switch to the release binary** — download the matching release asset for this OS/arch from the v<latest> GitHub release, **verify its checksum** against `detent_<ver>_checksums.txt`, then atomically replace the binary **at the detected path** (the `go install` one), with a clear **warning** that this changes how the binary is managed (future `go install`/`go.mod` upgrades won't track it; it's now release-pinned).
3. **Abort** — print the command and exit (current behavior, but only as an explicit choice).

`--yes` should pick a sane default non-interactively (recommend: **run the go install**, option 1 — least surprising for a go-install binary) and document it. Add a flag to force the release-swap path non-interactively (e.g. `--from-release`).

## Safety requirements (best practice for self-replacing binaries)

- **Checksum-verify** every downloaded asset before installing (the release ships `checksums.txt`).
- **Atomic install**: download to a temp file in the same dir, verify, `chmod`, then `os.Rename` over the target (rename is atomic on same filesystem); never partially overwrite the running binary.
- **macOS code signature**: a naively-copied/replaced Go binary on macOS gets **SIGKILLed** (exit 137) because the ad-hoc signature breaks — re-sign (`codesign -s -`) or ensure the release artifact carries a valid signature, and **verify the replaced binary runs** (`detent version` exit 0) before declaring success; roll back the prior binary if it doesn't. (We hit exactly this SIGKILL during ops.)
- **Permissions** preserved (0755).
- **Running process**: note that replacing the on-disk binary doesn't change the already-running process — instruct the user to restart (or, for the orchestrator, that the running instance keeps the old version until restarted).
- Clear **warning** before the release-swap path (irreversible management change).

## References (verify/cite current)
- go-selfupdate (github.com/sanbornm/go-selfupdate), blockthrough/selfupdate.go — established patterns for download+verify+atomic-replace from GitHub Releases.
- One2N "Making your Go CLI binaries self-updatable" — the go-install vs release-binary distinction.

## Acceptance
- `detent update` (interactive) on a go-install binary offers: run-go-install / switch-to-release(+warning) / abort — and executes the chosen one.
- `detent update --yes` no longer exits non-zero on a go-install binary; it performs the documented default (run go install) and reports the new version. `--from-release` forces the release-swap path.
- Release-swap path: checksum-verified, atomic, signature-valid (binary runs post-replace, with rollback on failure), perms preserved, with a warning.
- `make check` green; tests for source detection + the decision branches + checksum/atomic-replace logic.

corylanou:
## Summary

- add go-install update choices for run-go-install, release swap, and abort
- add `--from-release` for explicit release-asset swaps from go-installed binaries
- verify release swaps with checksum, atomic replace, preserved permissions, macOS signing, post-install version checks, and rollback on verification failure
- document the go-install `--yes` default and release-pinned behavior

Fixes #317

## Test Plan

- [x] `go test ./internal/update ./cmd/detent`
- [x] `make check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,8 +170,14 @@ Release-installer installs can update with `detent update`; use
 `detent update --yes` for non-interactive automation and `detent update --json`
 for machine-readable status. On Windows, replacement is staged and completes
 after the running `detent.exe` exits. Homebrew installs delegate to
-`brew upgrade digitaldrywood/tap/detent`, while Go and source builds print the
-recommended command instead of overwriting the binary.
+`brew upgrade digitaldrywood/tap/detent`. Go-installed binaries offer an
+interactive choice: run
+`go install github.com/digitaldrywood/detent/cmd/detent@latest`, switch to the
+checksum-verified release binary, or abort. `detent update --yes` runs the Go
+install command for go-installed binaries; `detent update --from-release`
+switches the detected Go-installed binary to the release asset and pins future
+updates to release-binary management. Source builds still print the recommended
+command instead of overwriting the binary.
 
 ## Release
 
--- a/cmd/detent/update.go
+++ b/cmd/detent/update.go
@@ -26,6 +26,7 @@ type updateFactory func() (updateRunner, error)
 func newUpdateCommand(factory updateFactory) *cobra.Command {
 	var checkOnly bool
 	var assumeYes bool
+	var fromRelease bool
 	var jsonOutput bool
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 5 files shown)


corylanou:
@codex review

PR #336, opened 2026-06-09
fix(codex): unblock transport read loop

### LINKED ISSUE #326: fix(codex): readLoop back-pressure deadlock leaks goroutine + OS process group on turn error
## Problem
`internal/codex/transport.go` can **leak two goroutines and an OS process group** when a turn errors while the codex child is still streaming.

`readLoop` (`:224-237`) writes frames into the buffered `received` channel (cap 64). If the consumer stops calling `Receive` — e.g. `awaitResponse`/`streamTurn` returns an error and `RunTurn` returns early — while codex keeps emitting, the buffer fills and `readLoop` wedges forever on `t.received <- …` and never closes `readDone`. `wait()` is then stuck on `<-t.readDone`, so `cmd.Wait()` never runs and `t.done` never closes. `Close()`’s `<-t.done` path can't fire (even the post-kill 1s ctx-timeout fallback won't unblock because `readDone` stays open) → both goroutines plus the process-group reap leak. This compounds over a long-running supervisor.

## Fix
- Have `readLoop` also select on a quit/`done` channel when sending to `received`, and have `Close`/`wait` close that quit channel so `readLoop` can drain-and-exit instead of blocking.
- Ensure `readDone` is always closed on the `readLoop` exit path so `wait()`/`Close()` make progress.

## Acceptance criteria
- A test that errors a turn mid-stream (codex still producing frames) shows `readLoop` + `wait` exit and the process group is reaped (no goroutine leak — `goleak` or equivalent).
- `Close()` returns promptly in that scenario.
- `make check` green.

corylanou:
## Summary
- Add a read-stop signal so `Close()` stops publishing into `received` when the consumer has returned, while `readLoop` keeps draining stdout until the child exits or is killed.
- Cover both failed-turn and successful-turn back-pressure paths with helper processes that keep writing after the turn result, then assert `readLoop`, `wait`, and process reaping complete.

Fixes #326

## Test Plan
- `go test ./internal/codex -run "TestLocalTransportClose(ExitsAfterTurnErrorBackpressure|DrainsAfterSuccessfulTurnBackpressure)" -count=1`
- `go test -race ./internal/codex -run "TestLocalTransportClose(ExitsAfterTurnErrorBackpressure|DrainsAfterSuccessfulTurnBackpressure)" -count=1`
- `go test ./internal/codex -count=1`
- `go test -race ./internal/codex -count=1`
- `make check`

### DIFF
--- a/internal/codex/transport.go
+++ b/internal/codex/transport.go
@@ -34,11 +34,13 @@ type localTransport struct {
 	stdin          io.WriteCloser
 	codec          *Codec
 	received       chan transportResult
+	readStop       chan struct{}
 	readDone       chan struct{}
 	done           chan struct{}
 	sendLock       chan struct{}
 	waitErr        error
 	waitMu         sync.Mutex
+	readStopOnce   sync.Once
 	closeOnce      sync.Once
 	closeErr       error
 }
@@ -87,6 +89,7 @@ func (f *LocalTransportFactory) NewTransport(ctx context.Context) (Transport, er
 		stdin:          stdin,
 		codec:          NewCodec(stdout, stdin),
 		received:       make(chan transportResult, 64),
+		readStop:       make(chan struct{}),
 		readDone:       make(chan struct{}),
 		done:           make(chan struct{}),
 		sendLock:       make(chan struct{}, 1),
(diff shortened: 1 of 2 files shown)


corylanou:
@codex review

Issue #337, opened 2026-06-09
release(update): provision minisign checksum signing key


corylanou:
Depends on: #324

## Problem
Issue #324 makes release self-update fail closed unless the release checksum file has a detached minisign signature and the Detent binary has the pinned minisign public key for that release stream. The current release workflow publishes archives and checksums only, and the project release signing key has not been provisioned in the repo.

## Work
- Generate and store the Detent release minisign secret key as a human-owned GitHub Actions secret.
- Pin the matching minisign public key for self-update verification.
- Update the GoReleaser release workflow to publish `<checksum>.minisig` beside the checksum file.
- Verify a snapshot/release run emits the checksum signature asset and that `detent update` accepts a valid signature.

## Blocker
Human action is needed to create and own the release signing key and provide the GitHub Actions secret.

Follow-up from #324.

corylanou:
## Provisioning complete — ready to implement

The release signing key has been generated and the GitHub Actions secrets are set. Remaining work is code/CI only (no further human action).

### Pinned public key
Set `defaultChecksumMinisignPublicKey` in `internal/update` to the minisign **base64 key line** (the parser `parseMinisignPublicKey` expects the 42-byte base64 packet, i.e. the second line of the `.pub`):

```
RWR9LKG/4dIAmbeS4Ow7XOSjxOE7GKiwQ0OPHNoViW5V0FSo4WU0vucx
```

Full public key for reference (key ID `9900D2E1BFA12C7D`):
```
untrusted comment: minisign public key 9900D2E1BFA12C7D
RWR9LKG/4dIAmbeS4Ow7XOSjxOE7GKiwQ0OPHNoViW5V0FSo4WU0vucx
```

### GitHub Actions secrets (already set on the repo)
- `MINISIGN_KEY` — the minisign secret-key file contents
- `MINISIGN_PASSWORD` — the key password

### Implementation
1. Pin the public key above into `defaultChecksumMinisignPublicKey` (flips enforcement on — see #324/#338).
2. Update the GoReleaser release workflow to sign `checksums.txt` and publish `checksums.txt.minisig` beside it, using `MINISIGN_KEY` + `MINISIGN_PASSWORD` (write the key to a temp file in CI, pipe the password to `minisign -S`; clean up the temp key).
3. Verify a snapshot/release run emits `checksums.txt.minisig` and that `detent update` accepts a validly-signed release and rejects a tampered checksum.

### Acceptance criteria
[...]

PR #338, opened 2026-06-09
fix(update): harden self-update trust

### LINKED ISSUE #324: security(update): harden self-update — download bounds + checksum signature verification + preserve codesign
## Problem
The self-update path has two hardening gaps (`internal/update/update.go`):

1. **No transport bounds (DoS / OOM).** The HTTP client defaults to `http.DefaultClient` (`:108`) which has **no timeout**, and `Download` does `io.ReadAll(resp.Body)` (`:153`) with no size cap — the entire release binary is buffered in RAM. The CLI passes `cmd.Context()` with no deadline, so nothing bounds a hung or hostile-but-TLS-valid mirror.
2. **Integrity is TLS-only.** `checksums.txt` is fetched over the same channel as the archive and is itself **unsigned** — there is no GPG/cosign/minisign verification anywhere in the package. The verified-by-checksum binary is then `os.Rename`d over the **running executable** and ad-hoc signed on macOS via `codesign -s -` (`:943`), which also **strips any Developer-ID/notarization** the asset shipped with.

## Fix
- Set a default `&http.Client{Timeout: …}` in `NewGitHubClient`; wrap the body in `io.LimitReader` with a sane max archive size.
- Verify a detached signature on `checksums.txt` (cosign or minisign) against a **pinned public key** before trusting any checksum. If signing infra isn't ready yet, at minimum document that integrity depends entirely on GitHub TLS and gate the rest behind a follow-up.
- Only re-sign on macOS if the asset is unsigned; otherwise preserve the shipped signature (don't silently downgrade to ad-hoc).

## Acceptance criteria
- Update download is bounded by both a client timeout and a max-size `LimitReader`; tests cover timeout + oversize rejection.
- A tampered `checksums.txt` (bad signature) aborts the update before any file replacement.
- macOS path no longer strips a valid Developer-ID signature (or the downgrade is explicitly documented and opt-in).
- `make check` green.

corylanou:
## Summary

- Bound self-update HTTP downloads with a default timeout and max-size reader.
- Require minisign-signed checksum assets before trusting archive checksums; fail closed until release key/signature publication is provisioned in #337.
- Preserve valid macOS code signatures and only ad-hoc sign unsigned binaries.

Fixes #324

## Test Plan

- [x] `go test ./internal/update`
- [x] `make check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,13 @@ switches the detected Go-installed binary to the release asset and pins future
 updates to release-binary management. Source builds still print the recommended
 command instead of overwriting the binary.
 
+Release self-updates verify SHA256 checksums fetched from GitHub releases. The
+checksum verifier supports detached minisign signature assets named
+`<checksum>.minisig`, but enforcement is gated until the binary embeds the
+pinned minisign public key for the release stream. Until that release signing
+key is provisioned in #337, update integrity still depends on GitHub TLS plus
+the published checksum file.
+
 ## Release
 
 Cut releases from `main` by pushing a semver tag:
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/lmittmann/tint v1.1.3
 	github.com/pressly/goose/v3 v3.27.1
 	github.com/spf13/cobra v1.10.2
+	golang.org/x/crypto v0.50.0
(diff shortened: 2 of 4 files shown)


corylanou:
@codex review

PR #341, opened 2026-06-09
perf(web): cache snapshot enrichment

### LINKED ISSUE #329: perf(web): enrich snapshot once per publish instead of per SSE connection (N× store queries)
## Problem
Snapshot enrichment is recomputed **per SSE connection per tick** instead of once per published snapshot (`internal/web`).

- `internal/web/sse.go:59-63` — on every snapshot tick, each connected client independently runs `enrichSnapshot` → `snapshotCycleTime` + `snapshotBudget`, both of which **hit the store** (`CycleTimeReport`, `BudgetCostEvents`). With N clients that's N× identical store queries per tick.
- `internal/web/server.go:211-226` — `latestSnapshot` calls `hub.Subscribe(ctx)` (allocates a channel, registers in the subscriber map, may spawn a ctx-watch goroutine, then immediately `Close()`s) on **every** dashboard HTTP render, purely to read the latest value — when `hub.Latest()` already exists and is used in `api.go:29`.

## Fix
- Enrich once per published snapshot (at publish time, or cache the enriched result keyed to the snapshot) and fan the enriched value out to all subscribers.
- Replace `latestSnapshot`’s subscribe/close dance with `snapshot, _ := s.hub.Latest()` + `enrichSnapshot`.

## Acceptance criteria
- Store query count per snapshot tick is independent of the number of connected SSE clients (test or instrumentation).
- Dashboard render no longer creates+destroys a hub subscription per request.
- `make check` green.

corylanou:
## Summary

- Cache web snapshot enrichment per snapshot so SSE subscribers and API/dashboard readers reuse the same store-backed cycle-time and budget data.
- Read the dashboard snapshot with `hub.Latest()` instead of creating a temporary subscription.
- Add regression tests for SSE enrichment query counts, dashboard latest reads, and canceled-context cache poisoning.

Fixes #329

## Test Plan

- [x] `go test ./internal/web`
- [x] `make check`

### DIFF
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -30,7 +30,7 @@ func (s *Server) apiState(c echo.Context) error {
 	if !ok {
 		return c.JSON(http.StatusOK, snapshotErrorResponse(now, "snapshot_unavailable", "Snapshot unavailable"))
 	}
-	snapshot = s.enrichSnapshot(c.Request().Context(), snapshot)
+	snapshot = s.cachedEnrichedSnapshot(c.Request().Context(), snapshot)
 
 	return c.JSON(http.StatusOK, stateResponse(snapshot, generatedAt(snapshot, now)))
 }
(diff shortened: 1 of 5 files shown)


corylanou:
@codex review

PR #342, opened 2026-06-09
fix(runtime): harden retry subprocess waits

### LINKED ISSUE #331: fix(runner): cap supervisor backoff exponent (int64 overflow) + set WaitDelay on git/hook subprocesses
## Problem
Two subprocess/retry robustness issues in the agent runtime:

1. **Backoff integer overflow** (`internal/runner/supervisor.go:144`). `failureRetryBaseDelay * time.Duration(math.Pow(2, float64(exponent)))` with `exponent` capped at 30 and a 10s base yields ~1.07e19 ns, which **overflows int64 `time.Duration`** (max ~9.22e18) and becomes negative; the `delay > maxRetryBackoff` guard then fails and a **negative delay** is returned. Reachable after ~31 retries.
2. **Missing `WaitDelay` on subprocesses** (`internal/workspace/workspace.go` `runGitAtWithEnv` `:525`, hooks `:489`, diffstat git calls). `exec.CommandContext` kills on ctx cancel, but with no `cmd.WaitDelay` a git child blocked on a credential/SSH prompt can hang `CombinedOutput()` indefinitely despite cancellation.

## Fix
- Clamp the backoff exponent so the product can't overflow (or early-return `maxRetryBackoff` once the exponent exceeds a safe bound).
- Set `cmd.WaitDelay` on the git/hook `exec.Cmd` values so cancellation force-kills and the call returns promptly.

## Acceptance criteria
- Backoff never returns a negative or overflowed delay at any attempt count (table test).
- Cancelling a wedged git/hook subprocess returns within `WaitDelay` (test or documented).
- `make check` green.

corylanou:
## Summary

- Clamp supervisor retry backoff multiplication before it can overflow `time.Duration`.
- Set `WaitDelay` on workspace git and hook subprocesses so cancelled commands return promptly when descendants keep pipes open.
- Preserve successful daemonized hooks that hit `exec.ErrWaitDelay` without cancellation.
- Add regression coverage for large retry attempts and wedged git/hook subprocess cancellation.

Fixes #331

## Test Plan

- [x] `go test ./internal/runner -run TestSupervisorRetryDelayNeverOverflows`
- [x] `go test ./internal/workspace -run 'TestRunGitAtWithEnvCancellationReturnsPromptly|TestLocalGitHookCancellationReturnsPromptly|TestLocalGitHookAllowsDaemonizedSuccess'`
- [x] `go test ./internal/runner ./internal/workspace`
- [x] `make check`

### DIFF
--- a/internal/runner/supervisor.go
+++ b/internal/runner/supervisor.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"math"
 	"sync"
 	"time"
 )
@@ -136,12 +135,14 @@ func (s *Supervisor) RetryDelay(attempt int) time.Duration {
 	if attempt < 1 {
 		attempt = 1
 	}
-	exponent := attempt - 1
-	if exponent > 30 {
-		exponent = 30
-	}
 
-	delay := failureRetryBaseDelay * time.Duration(math.Pow(2, float64(exponent)))
+	delay := failureRetryBaseDelay
+	for range attempt - 1 {
+		if delay >= maxRetryBackoff || delay > maxRetryBackoff/2 {
+			return maxRetryBackoff
+		}
+		delay *= 2
+	}
 	if delay > maxRetryBackoff {
 		return maxRetryBackoff
 	}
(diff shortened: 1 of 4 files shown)


corylanou:
@codex review

PR #1008, opened 2026-07-08
fix(kanban): replace native HTML5 drag with pointer-event dragging

loganlanou:
## Summary

- **Root cause of the dead drags:** Chromium on Wayland opens a compositor-level drag session even for drags that never leave the page, and several compositors — including COSMIC (cosmic-comp, what our Pop!_OS machines run), Plasma, and Hyprland — cancel or break that session (Chromium's "[ozone/wayland] Drag'n drop breaks", https://issues.chromium.org/issues/40707943; COSMIC DnD issues, https://github.com/pop-os/cosmic-epoch/issues/1175). A real mouse drag dies the instant it starts, so cards can never be picked up. Playwright and CDP `dispatchDragEvent` drags bypass the native session entirely, which is why #998, #1004, #1005, and #1007 all shipped with green drag tests while human drags stayed dead.
- **Fix:** stop using native HTML5 drag and drop — the same conclusion the ecosystem reached (dnd-kit, hello-pangea/dnd and other kanban libraries track pointers instead of wrapping the native API). Card dragging is now pointer-event based: `pointerdown` arms a candidate, a 6px movement threshold starts the drag, a cloned ghost follows the pointer, `elementFromPoint` hit-tests the drop lane, and `pointerup` submits the existing move form. Nothing touches the compositor, so it works on every platform.
- Details preserved from the old flow: allowed/blocked lane states and feedback messages are unchanged, drops still post `/api/v1/kanban/move` via the per-card hidden form, morph-refresh mid-drag still re-asserts drag state, read-only cards stay inert. New: Escape / `pointercancel` / window-blur cancel cleanly, the click that trails a drag is swallowed so the detail sheet doesn't open, the lane strip auto-scrolls at its edges (native DnD used to do this for free), and native drags inside cards are suppressed so links/images can't start a compositor drag that steals the pointer.
- **The test blind spot is closed:** Playwright's `page.mouse` emits real pointer events, so the drag tests now exercise the exact production code path. Replaced the obsolete dragstart-deferral regression test with a pointer-path test (threshold behavior, drag affordances, Escape cancel, no stray move request, no sheet opening).

## UI Surface Contract

## Test Plan

- [x] `make check` equivalent: build, vet, nilaway, race tests, 75.3% coverage vs 70% gate (local `make lint` still reports the 57 pre-existing findings that reproduce on clean `main`; CI Lint is the authoritative gate)
- [x] `make visual-e2e` — all 26 Playwright tests pass; the drag tests now drive the real pointer path
- [x] Browser verification against `detent dev-runtime --demo kanban`: pointerdown alone does nothing, sub-threshold movement does nothing, past-threshold movement shows the ghost + unhides lanes with allowed/blocked states + "Moving Backlog." feedback, dropping on Todo posts the move and the card lands in the Todo lane, drop cleanup restores hidden lanes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


NicTanghe:
I also cant drop in x11 using COSMIC browser

PR #1394, opened 2026-07-17
fix(web): unblock board page rendering

### LINKED ISSUE #1390: bug(web): project board pages block ~5s server-side while / renders in 1ms — fixed timeout on the render path
## Problem

Project board pages render ~5 seconds server-side while the main board renders in ~1ms. Measured 2026-07-17 on mac-studio (v0.44.0, steady state, fully hydrated, zero running sessions), repeatable:

```
/                     0.0013s
/projects/detent      5.07s
/projects/gopher-ai   5.03s
/fleet                2.42s
/health               0.0004s
```

Both project pages landing at almost exactly 5.0s suggests a fixed internal timeout/budget being waited out on every request (per-request tracker hydration or a probe with a 5s deadline) rather than genuine work — the main board serves the same underlying data in ~1ms from the hub snapshot. `/fleet` at 2.4s likely shares a smaller variant of the same cost.

This is the surviving half of the "board-to-board navigation sometimes stalls" complaint from #1356 (closed): the post-restart refresh work shipped there, but project-page navigation still blocks.

## Task

- Instrument the project-page handler to attribute the ~5s (which call, which timeout).
- Serve project pages from the same already-hydrated snapshot the main board uses; no per-request tracker round-trips or fixed waits on the render path.
- Same for `/fleet` if it shares the cause.

## Acceptance criteria
- [ ] Root cause named in the PR (the specific call/timeout consuming the ~5s).
- [ ] `/projects/*` and `/fleet` server render times measured before/after in the PR; after must be comparable to `/` (well under 500ms on a hydrated instance).
- [ ] Regression test covering the render path not performing blocking tracker calls.
- [ ] New Playwright spec asserts board-to-board navigation completes without the multi-second stall.
- [ ] Desktop visual baselines unchanged.
- [ ] `make check` passes.

```detent-agent
schema: 1
effort: high
```

Related to #1356 (closed: post-restart refresh; this is the navigation remainder) and #1318 (first-paint hub snapshot pattern to reuse).

corylanou:
## Summary

- Render fleet and every project page from the hub's already-hydrated in-memory snapshot, leaving store-backed enrichment on the existing SSE path instead of the HTTP first-paint path.
- Add a blocking-store regression matrix for fleet, project overview, Kanban, runs, and diagnostics, plus a Playwright board-to-board navigation test.
- Root cause: `/fleet` and non-Kanban project handlers called `latestSnapshot`, which enters synchronous `snapshotEnrichmentCache.enrich`; an in-flight enrichment waits on runtime-store queries beginning with `CycleTimeReport`, and the runtime SQLite connection's default `PRAGMA busy_timeout` is 5000ms. `/` and project Kanban already avoided that wait through `latestBoardSnapshot`.
- Before on the hydrated mac-studio instance: `/` 1.3ms, `/projects/detent` 5.07s, `/projects/gopher-ai` 5.03s, `/fleet` 2.42s. After in an isolated hydrated runtime (10 requests, p50): `/` 3.44ms, `/projects/demo-project` 1.20ms, `/projects/demo-project/kanban` 0.95ms, `/fleet` 3.30ms.

Fixes #1390

## UI Surface Contract

## Test Plan

- [x] `make check`
- [x] `go test ./internal/web -count=1`
- [x] `playwright test tests/visual/board-navigation.spec.js --project=chromium`
- [x] `playwright test tests/visual/layout.spec.js --project=chromium` (41 passed; desktop baselines unchanged)

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -396,7 +396,8 @@ func (s *Server) dashboard(c echo.Context) error {
 		return s.demoDashboard(c, scenario)
 	}
 	ctx := c.Request().Context()
-	data := s.dashboardData(ctx, s.latestSnapshot(ctx))
+	snapshot, enriched := s.latestBoardSnapshot()
+	data := s.dashboardFirstPaintData(ctx, snapshot, !enriched)
 	applyDashboardPreferences(c.Request(), &data)
 	return render(c, templates.FleetPage(data))
 }
@@ -512,14 +513,8 @@ func (s *Server) projectDashboard(c echo.Context) error {
 		}
 		return s.demoProjectDashboard(c, scenario, view)
 	}
-	var data templates.DashboardData
-	var ok bool
-	if view == "kanban" {
-		snapshot, enriched := s.latestBoardSnapshot()
-		data, ok = s.projectBoardFirstPaintData(ctx, projectID, snapshot, !enriched)
-	} else {
-		data, ok = s.projectDashboardData(ctx, projectID, s.latestSnapshot(ctx))
-	}
+	snapshot, enriched := s.latestBoardSnapshot()
+	data, ok := s.projectFirstPaintData(ctx, projectID, snapshot, !enriched)
 	if !ok {
 		return echo.NewHTTPError(http.StatusNotFound, "Project not found")
 	}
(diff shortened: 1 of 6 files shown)


corylanou:
Addressed the automated P2 in adb8af51: budget-enabled project pages now render an explicit pending state for the budget panel and spend metrics until SSE enrichment arrives, instead of showing zero spend or omitting an active override. The blocking-store regression asserts the pending state and absence of the misleading $0.00 / $100.00 value.

PR #1494, opened 2026-07-24
fix(doctor): freeze parent-cancel progress deterministically

### LINKED ISSUE #1493: fix(doctor): freeze parent-cancel progress deterministically
## Summary

PR #1492 fixed liveness-timeout snapshots by freezing completed checks and the current sub-check before its explicit timer cancellation. Final independent review found two cancellation-boundary gaps that remain after that merge.

## Parent cancellation race

`runDoctorCheck` derives `checkCtx` directly from the parent context. When the parent is canceled, `checkCtx` is already canceled before the `<-ctx.Done()` branch can call `Freeze`. A context-aware sub-check can wake, publish a cancellation-derived result, and advance its marker before `Freeze` acquires the progress lock. The captured partial report is therefore scheduler-dependent.

Derive the worker context from `context.WithoutCancel(ctx)`, wrap it with `context.WithCancel`, then atomically freeze progress before explicitly canceling the worker context when the parent ends. Preserve the parent error in the synthesized failure.

## Regression test gap

`TestRunDoctorCheckFreezesProgressBeforeCancellation` does not deterministically fail if cancel-before-`Freeze` is reintroduced. The caller can still win the mutex race and capture the old snapshot. Its `Current` callback cannot enforce ordering because `freezeDoctorCheck` bypasses `Current` when `Freeze` is configured.

Wrap the test `Freeze` callback with an assertion that the worker context is not canceled at snapshot time, or add an equivalent synchronization hook. Cover both timer expiration and parent-context cancellation.

## Acceptance criteria

- Timer and parent-cancellation paths freeze completed results and the current stage before the worker sees cancellation.
- Post-freeze publications remain rejected.
- Deterministic tests fail if either path cancels before freezing.
- Existing timeout and progress tests remain green under the race detector.
- `make check` passes.

Depends on: none

```detent-agent
schema: 1
effort: xhigh
```

References: #1489 and PR #1492.

corylanou:
## Summary

- Isolate doctor worker cancellation from parent cancellation so progress freezes before the worker observes cancellation.
- Add deterministic timer-expiration and parent-cancellation ordering coverage, including post-freeze publication rejection.

Fixes #1493

## UI Surface Contract

- [x] N/A; this PR changes doctor cancellation ordering and tests only.
- [x] N/A; this PR has no visual changes.

## Test Plan

- [x] `go test -race ./internal/cli -run TestRunDoctorCheck -count=20`
- [x] `make check`

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -632,7 +632,7 @@ func runDoctorCheck(ctx context.Context, job doctorCheckJob, timeout time.Durati
 		ctx = context.Background()
 	}
 	timeout = doctorNormalizedTimeout(timeout)
-	checkCtx, cancel := context.WithCancel(ctx)
+	checkCtx, cancel := context.WithCancel(context.WithoutCancel(ctx))
 	defer cancel()
 	timer := time.NewTimer(timeout)
 	defer timer.Stop()
--- a/internal/cli/doctor_test.go
+++ b/internal/cli/doctor_test.go
@@ -4723,40 +4723,89 @@ func TestRunDoctorCheckRenewsTimeoutForReportedProgress(t *testing.T) {
 func TestRunDoctorCheckFreezesProgressBeforeCancellation(t *testing.T) {
 	t.Parallel()
 
-	progress := newDoctorCheckProgress()
-	postCancelPublication := make(chan struct{})
-	completed := doctorCheck{Name: "Project alpha workflow", Status: doctorOK, Detail: "loaded"}
-	canceled := doctorCheck{Name: "Project alpha route models", Status: doctorFail, Detail: "context canceled"}
-	checks := runDoctorCheck(context.Background(), doctorCheckJob{
-		Name:   "Project alpha checks",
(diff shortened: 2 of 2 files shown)


corylanou:
One regression gap remains. `TestRunDoctorCheckFreezesProgressBeforeCancellation` attempts a post-freeze `Set`, but every assertion inspects the snapshot that `Freeze` already copied. If `Set` stopped honoring the frozen guard, the test would still pass because the returned checks cannot change. After `postCancelPublication` closes, inspect the progress object again and assert its current stage and completed checks remain at the frozen values, or add a focused `Set`-after-`Freeze` test. This directly covers the acceptance criterion that post-freeze publications remain rejected.

corylanou:
Addressed in a61499ed. After the worker signals its post-cancellation publication attempt, the test snapshots the progress object again and asserts both the current stage and completed checks remain at the frozen values. A mutation run with the frozen guard removed failed both timer and parent-cancellation subtests 5/5 at the new assertion. Focused race tests and `make check` pass.

PR #2502, opened 2026-09-11
fix(doctor): scope INV-4 evidence to 24h and exempt push-only jobs from INV-5

corylanou:
## Summary

- INV-4 evidence now counts programmatic merges in the last 24h, not 7d: a queue enabled recently otherwise reports every pre-queue merge as a bypass (the dogfood project showed 248).
- INV-5 treats a job whose `if` names only other events (`github.event_name == 'push' || … 'workflow_dispatch'`) as off pull_request; this repository's integration jobs were misreported as running on every PR push.
- Found by running the v0.114.0 candidate's `detent doctor` against the live mac-studio store before tagging.

Invariants touched: INV-4, INV-5 (doctor evidence windows only; docs/invariants.md statements unchanged).

## UI Surface Contract

## Test Plan

- [x] `go test ./internal/cli -run 'TestDoctorInvariant|TestCheckDoctorProjects'`, `make lint`
- [x] New cases: programmatic merges older than 24h are not INV-4 evidence; push-only integration jobs pass INV-5

https://claude.ai/code/session_01PK61Vw2Gr7EMRGtKohPXrw

### DIFF
--- a/docs/invariants.md
+++ b/docs/invariants.md
@@ -86,6 +86,13 @@ exercises native queue delegation. `TestRepositoryWorkflow` requires
 their chosen settings; Detent detects the queue and uses its serialized merge
 worker where no queue exists. These tests do not inspect live GitHub settings.
 
+`detent doctor` reports runtime evidence: with a queue present, any
+`merge_worker_programmatic_merge` lane event or `programmatic_merge_failed`
+attempt in the last 24 hours fails the check. The store records no queue
+activation time, so a merge from before a queue was enabled can fail the
+check until it ages out of that day; a second queue removal routing to Human
+Review is the budget, not a bypass.
+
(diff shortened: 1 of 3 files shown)


corylanou:
Addressed: docs/invariants.md INV-4 and INV-5 now describe the doctor evidence windows and the static workflow verdict. INV-5 exemption now requires every || alternative to carry a github.event_name equality with no 'pull_request' anywhere (github.event_name == 'push' || github.event.action == 'opened' is a real job; test added). INV-4 stays a 24h window: the store records no queue activation time (the only durable marker is the 405 rule-violation error, which Detent no longer produces once it respects the queue), so the doc states that a pre-queue merge can fail the check until it ages out of that day rather than pretending to anchor on data that does not exist.

PR #2503, opened 2026-09-11
fix(boot): continue startup when prior-instance worker cleanup fails

corylanou:
## Summary

- `reapWorkerProcesses` at startup is now best effort: its error is logged as a warning and boot continues. Every start of v0.114.0 on mac-studio aborted with `reap worker processes from prior instance: … workspace scan command … context deadline exceeded` because six dead detent.build#50 worker sessions pointed at a 15 GB retained workspace whose `lsof +D` scan cannot finish inside the deadline. The prior instance's processes were already gone; refusing to serve the board over a cleanup timeout is the wrong failure mode.

Invariants touched: none (INV-2 spirit: an instance-level cleanup failure no longer takes the service down).

## UI Surface Contract

## Test Plan

- [x] `go test ./internal/cli -run 'Boot|Reap|WorkerProcess'`, `make lint`
- [x] Reproduced on mac-studio: two consecutive starts failed identically; after marking the dead sessions reaped the service starts

https://claude.ai/code/session_01PK61Vw2Gr7EMRGtKohPXrw

### DIFF
--- a/internal/cli/boot.go
+++ b/internal/cli/boot.go
@@ -263,10 +263,14 @@ func startRunningWithDependencies(ctx context.Context, cfg BootConfig, deps star
 			logShutdownBoundaryEnd(logger, "runtime_store_close", closeStarted, nil, "component", "runtime_store")
 		}
 	}()
-	// Prior-instance cleanup is best effort: a workspace scan that times out
-	// on a large tree must not keep this instance from serving.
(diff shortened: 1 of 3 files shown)


corylanou:
Addressed: reapWorkerProcessesWithCleanup now returns a typed workerArtifactCleanupError only when every prior worker was confirmed terminated (or already exited) and just the artifact cleanup failed; boot continues in that case and keeps the session rows unreaped for retry. Any termination failure (EPERM, inspection error, alive after SIGKILL) or reap-marking failure is still fatal at startup. TestReapWorkerProcessesClassifiesCleanupOnlyFailures covers the three combinations.

PR #2508, opened 2026-09-11
fix(orchestrator): track validator sessions in runtime snapshots

### LINKED ISSUE #2505: fix(orchestrator): worker session started 27s after boot is untracked in Running and flagged as an orphan
## Problem

After the v0.114.1 restart on mac-studio (process start 14:54:07 CDT), a Codex worker session for digitaldrywood/detent#2486 (codex_sessions id 6059, created 19:54:12Z, `worker_session_started` at 19:54:39Z, pid 80484, ppid = the new detent process) ran for several minutes while:

- `/api/v1/state` reported `counts.running: 0`, an empty `running` array, and no object carrying `detent_session_id` anywhere in the document;
- `/health` reported `agent_memory: []` and `orphaned_agent_processes.count: 2` (`session_count: 1`, session 6059), which set `status: needs_attention`;
- `workflow_phase_events` has `agent_active` entries for the two earlier #2486 sessions (6056 at 19:39:48Z, 6058 at 19:49:02Z, the latter killed by the restart) but no lane event for 6059;
- the log shows `worker_agent_update` events for session 6059 streaming (the agent was working) but no non-DEBUG dispatch/claim line for it, and `project startup recovery timing` lines at 19:54:11–19:54:13Z immediately before.

So the instance spawned an agent it does not count as running. Whatever path started 6059 (startup recovery of the session the restart killed, or an early dispatch before the first refresh) must register the session in `Running` like a normal dispatch, or not start it. Health then flags the instance's own child as orphaned, which is the correct symptom for an untracked agent but the wrong root cause to hand a human.

Operator action taken: SIGTERM to pgid 80484 at ~19:58Z so the untracked agent stops consuming the session budget; #2486 stays in Rework for a normal dispatch.

## Acceptance

- A test that starts an instance with a `running` codex session whose worker died (as the restart does) and asserts that any session the recovery path starts appears in the snapshot's `Running` and in `agent_memory`, and is not reported by `orphaned_agent_processes`.
- No new mechanism: fix the registration in the existing recovery/dispatch path (INV-3).

```detent-agent
schema: 1
effort: medium
```

Effort medium: the reasoning difficulty is which startup path created the session and why `Running` was bypassed, not file count.

corylanou:
## Summary

- Session 6059 was a validator with no work-attempt link. Validator launches were tracked only by identity, so startup validation was absent from Running and health classified the live worker as orphaned.
- Retain completed validator tokens on both success and error. Prefer implementation identity with deterministic validator fallback when sessions overlap.
- Publish validator session identity, memory, and progress through the existing validator lifecycle registry and shared worker progress tracking. Preserve concurrent implementation sessions and frozen completion observations; remove validators on exit. No new recovery mechanism or orphan-suppression guard.

Fixes #2505

Invariants touched: INV-3 — [same-PR consolidation documentation](docs/invariants.md#inv-3--mechanism-moratorium).

## UI Surface Contract

- [x] N/A: this issue authorizes correcting Running and health accounting; no layout or visibility tradeoffs.
- [x] Existing state and health HTTP handlers verified with startup-session fixtures; no visual layout changes.

## Test Plan

- [x] Missing validator registration reproduced before the fix.
- [x] Persisted running Codex session startup tests cover orphan resume and first-tick Rework validation, Running identity, agent_memory, and absence from orphaned_agent_processes.
- [x] Rework regression tests reproduce and fix completed-token loss and nondeterministic identity.
- [x] Validator success/error cleanup, concurrent implementation, snapshot copy isolation, and cached completion observations tested.
- [x] Existing startup/completion regressions, including frozen persisted heartbeats, pass.
- [x] `make check-fast` (configured worker gate; includes invariant gate).
- [x] Independent validator review: pass, 0.97, no P1/P2 findings.


corylanou:
Validator verdict: pass
- score: 0.96
- summary: Satisfies both acceptance criteria: startup regression tests verify session visibility in Running and agent_memory with zero reported orphans; validator accounting reuses the existing registry and progress publisher without introducing a new mechanism. Targeted regression tests and the invariant gate passed. Full make check was not rerun.

PR #2543, opened 2026-09-13
fix(runner): fall back when model catalog fails

### LINKED ISSUE #2540: fix(runner): automatic model selection fails every dispatch with 'model catalog unavailable' on one project
## Problem

On the second host, one project (a local work-item tracker project with `agent.model_selection` automatic policy) fails every dispatch within seconds: `run agent turn: automatic model selection: model catalog unavailable; retry catalog discovery before dispatch` (work_attempts at 15:56Z and 16:50Z on 2026-09-13 for the same work item, `runner_error`). `internal/runner/model_selection.go:167-169` returns this when the backend's `ListModels` fails, but the underlying error is discarded, so the operator cannot tell whether the Codex app-server model list call failed, the backend has no catalog provider, or the cache is stale. The same host runs explicit-model projects fine, so the catalog call itself is the suspect.

## Change

- Wrap and log the underlying `ListModels` error (`%w`) so the attempt record and log show the cause.
- If catalog discovery fails, fall back to the project's configured `normal_model` for that attempt instead of failing the dispatch (existing route defaults), and record the fallback. No new mechanism (INV-3).

## Acceptance

- Test: catalog error → attempt proceeds on the configured model with the reason recorded; catalog available → unchanged selection.
- `make check-fast` passes.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- fall back to the configured automatic-selection `normal_model` when backend catalog discovery fails
- preserve the originally requested model and record the wrapped catalog cause in attempt identity and structured logs
- cover selection, dispatch, persistence, logging, and error-classification behavior

Fixes #2540

Invariants touched: none. This uses the existing route/default selection and runtime identity provenance; no new mechanism is introduced (INV-3).

## UI Surface Contract

- [x] N/A; this PR has no UI changes.
- [x] N/A; this PR has no visual changes.

## Test Plan

- [x] `go test ./internal/runner -count=1`
- [x] `make check-fast`


corylanou:
Validator verdict: pass
- score: 0.98
- summary: Implementation satisfies the acceptance criteria: catalog failures fall back to the configured normal model under the existing fallback policy, preserve and record the diagnostic reason, log the failure, and leave catalog-available selection behavior unchanged. Error diagnostics are bounded and sanitized for backend response errors. make check-fast passed.

PR #2551, opened 2026-09-13
fix(github): scope REST budgets by endpoint

### LINKED ISSUE #2547: fix(github): REST reserve gate conflates rate-limit windows across endpoint families
## Problem

The REST reserve gate, the lookup-backoff floor, and the dispatch pause all read one `restRateLimits["core"]` snapshot per credential. That snapshot is overwritten by whichever response arrived last, and on 2026-09-13 one credential (`github-rest:14c4b783c5c9`) reported three different rate-limit windows at the same time, all labeled `core`:

| endpoint family | remaining | reset_at (UTC) | observed |
|---|---|---|---|
| check runs | 4721 | 19:03:56 | 18:19:16 |
| issue comments | 4723 | 19:03:56 | 18:19:20 |
| commit statuses | 1455 | 18:03:55 | 18:03:53 |
| issue dependencies / workflow runs | 314 | 18:19:28 | 18:19:16 (304) |

Same-cycle evidence (`github rest budget summary` at 18:19:23Z, `credential_budgets`): check runs at 4721 with reset 19:03:56 and workflow runs at 314 with reset 18:19:28, observed 200 ms apart. The reserve gate then refused a commit-status request using 314 (snapshot age 154 µs) while ordinary responses in the same second showed ~4720 remaining.

Impact on one host over ~80 minutes:

- 80 `github rest reserve floor held` refusals at remaining 300–1500 vs reserve 1000.
- scheduler reasons `github_rest_capacity_paused` and `pull_request_hydration_unavailable`, Merging items skipped.
- a second project on the same host went into `github lookup backoff` (`GitHub REST remaining 416 is at or below lookup floor 1000`, 4 consecutive probes) and skipped its entire candidate list (128 issues) with `github_lookup_recovery`; its snapshot aged to 81,711 s.

Mitigation applied by the operator: `tracker.github_rest_min_remaining_reserve: 300` on both projects. That hides the symptom; the accounting is still wrong.

## Verified against the code (independent review)

- Not the empty-resource fallback (`restRateLimitForResourceLocked`, client.go ~967): the reserve gate passes an explicit resource (~893) and the mapping never returns empty (~1740).
- Not a credential-hash collision: isolation uses the full hash (~1033), a key change clears state (~1057), and response updates require a matching key (~1098).
- Cause: the writer at client.go ~1153 unconditionally replaces `restRateLimits[resource]` with the accepted response, so responses carrying different windows (different `reset_at`) under the same `resource` label overwrite each other. The `resource` label comes from `X-RateLimit-Resource` when present and otherwise from the path mapping (~1081, ~1733), so a logged `core` does not prove GitHub said `core`. The same conflated snapshot feeds the summaries and the status surface (~792).

## Fix direction (moratorium-compliant)

Stop treating the last response labeled `core` as authority for every endpoint. The per-endpoint-family budget evidence already exists (`credential_budgets`); make the reserve gate, the lookup floor, and the dispatch pause decide from the family that the request belongs to, and consolidate the blanket `github_rest_capacity_paused` pause so an ambiguous aggregate cannot stop unrelated work. Preserve real 403/429 and `Retry-After` handling. Record the raw `X-RateLimit-Resource` value (or its absence) in the snapshot so the label is verifiable.

Do not fix by lengthening the credential hash, changing the empty-resource fallback, or picking the window with the largest remaining count.

## Acceptance

- A table-driven test where two responses for one credential carry different `reset_at` under the same resource label; the reserve gate for a check-runs request is not refused by the workflow-runs window.
- `github rest budget summary` no longer shows the same `resource` with differing `reset_at` inside one cycle, or it shows them under distinct keys.
- On the reporting host, zero `reserve floor held` at remaining > 2000 across a full core window after the default reserve (1000) is restored.

```detent-agent
schema: 1
effort: high
```

corylanou:
## Summary

- Isolate GitHub REST reserve and lookup decisions by credential and endpoint family so unrelated rate-limit windows cannot overwrite one another.
- Expose raw `X-RateLimit-Resource` provenance and retire the ambiguous aggregate connector bucket as a fleet-wide dispatch pause.
- Preserve explicit provider throttling and worker/shared-pool budget handling.

Fixes #2547

Invariants touched: [INV-3](https://github.com/digitaldrywood/detent/blob/detent/detent-digitaldrywood_detent_2547-52e7906b7d90/docs/invariants.md#inv-3--mechanism-moratorium)

## UI Surface Contract

- [x] N/A; this PR adds telemetry provenance only and does not change operator layout, density, first-viewport content, or responsive visibility.
- [x] N/A; there are no visual changes to primary operator screens.

## Test Plan

- [x] Table-driven regressions cover conflicting endpoint-family windows, header presence, and credential rotation.
- [x] Connector, orchestrator, and telemetry package suites pass.
- [x] `make check-fast`


corylanou:
Validator verdict: rework
- score: 0.58
- summary: The connector reserve gate and lookup floor are correctly family-scoped, but the PR leaves an acceptance-critical aggregate consumer that can still cancel work and recreate the fleet-wide dispatch pause.

Findings:
- Worker/shared-pool budgets are still accepted as authoritative capacity evidence here, although they originate from `/rate_limit.resources.core`; `workerGitHubPolicy.reserveError` also directly cancels running work from that aggregate. Current reporting-host evidence shows this probe reporting 1063/reset 21:19:46 while request-family telemetry simultaneously reported 4646/reset 22:03:59. Consequently the same ambiguous aggregate can still stop unrelated work and produce `github_rest_capacity_paused`, so the mechanism consolidation is incomplete. Remove/consolidate this aggregate enforcement or base it on relevant request-family evidence, with a regression test covering the worker/shared-pool divergence. (internal/orchestrator/github_rest_capacity.go:102)

corylanou:
Validator verdict: pass
- score: 0.93
- summary: The implementation satisfies the code-level acceptance criteria: REST reserve and lookup decisions are keyed by credential and endpoint family; same-resource responses with different reset windows remain distinct; check-runs admission is unaffected by a low workflow-runs window; raw X-RateLimit-Resource presence is reported; ambiguous aggregate telemetry no longer creates a blanket dispatch outage; and explicit backoff plus worker/shared-pool handling remain intact. Focused connector, orchestrator, and telemetry tests passed. The reporting-host full-window criterion remains an operational post-deployment observation.

Findings:
- The repository-wide make check reached the race suite but failed outside this diff when TestReapScratchProcessesOutsideWorkspace encountered a process-inspection context deadline and the workspace package timed out. The isolated failing race test passed immediately afterward, while the modified packages passed, indicating an environmental/flaky validation failure rather than a defect in this change. (internal/workspace/process_scratch_test.go:40)

PR #2705, opened 2026-09-14
fix(runner): inherit and bound host Go caches

### LINKED ISSUE #2680: fix(runner): drop the Go cache override; workers use the host toolchain cache, bounded by the reaper
Operator-approved scope 2026-09-14 (INV-11). This removes a mechanism; it adds no brake, breaker, or new loop.

## Evidence (mac-studio, 2026-09-14)

- The data volume hit 100% and Docker's VM shut down, taking all 12 self-hosted runners offline. Two Detent per-project Go caches held about 730 GB: `~/code/detent-workspaces/.detent/cache/detent-<sha12>` 567 GB and `~/code/parable-workspaces/.detent/cache/parable-<sha12>` 163 GB. Nothing in Detent trims them (no `GOCACHE`/`go clean`/size code outside `internal/runner/worker_cache.go:54`).
- `configureWorkerCache` (`internal/runner/worker_cache.go:18-62`) overrides `GOCACHE`, `GOMODCACHE`, `GOBIN` and `GOLANGCI_LINT_CACHE` for every worker. `isolated` (the default, `internal/config/config.go:1486`) points them at the per-attempt temp dir, so every attempt cold-compiles the whole dependency graph (#1437: golangci-lint alone 2,627 s, host load 60 on 12 cores). `shared` (#1454) points them at a root keyed only by project id (`worker_cache.go:66-74`), so each project keeps its own copy of the standard library, toolchain outputs and common deps, and the operator's own builds keep a third copy in the host cache (99 GB).
- Go already solves this: one content-addressed cache per user, documented as safe for concurrent go commands (`go help cache`), shared by every project and every checkout. The #1437 thread records that when a 0.46.0 regression dropped the override, workers fell through to the host cache and built fine.
- Churn measured before the operator's manual 48-hour trim: detent go-build 584 GB with 217 GB touched in the last day and 279 GB older than 2 days (1.81M files); parable 166 GB with 41% touched within 24 h. Go's own trim only drops entries unused for 5 days and has no size cap, so any cache, per project or host-wide, refills to hundreds of GB at this churn. The churn itself comes from the full-gate-per-session instruction pattern (#2659).
- Cross-project overlap measured by content-hash name: 1,997 entries, 0.64 GB shared between the detent and parable caches. Merging per-project caches into one saves nothing by itself; the value of dropping the override is one cache to bound, no Detent-owned cache directories, and no duplicate of the operator's own host cache for the same repos (not measured).

## Change

1. Delete the cache override. Workers inherit the host toolchain caches (`go env GOCACHE`, `GOMODCACHE`, `GOBIN` unset unless the operator sets it, `GOLANGCI_LINT_CACHE` unset). Keep `TMPDIR`/`TMP`/`TEMP` per attempt as today.
2. Sandbox: for `workspace-write` policies, add the resolved `GOCACHE` and `GOMODCACHE` directories to `ExtraWritableRoots` so builds work without full access.
3. Remove `workspace.cache_strategy`: config validation warns on the key as unknown-and-ignored for one release, then errors. Remove it from `docs/templates/*`, `config.reference.yaml`, docs/config.md and getting-started.md; the onboarding builder stops emitting it.
4. Migration: at boot, remove every legacy `<workspace.root>/.detent/cache/` root and log the bytes reclaimed. This is how prometheus (pyroapex cache 19 GB) cleans itself after the update; no manual purge.
5. Bounding, host-wide, inside the existing reaper sweep (`internal/orchestrator/reaper.go:50-111`), not a new loop: a `global.cache` block in global.yaml with `max_age` (default 48h) and `max_bytes` (default 1000 GiB, operator decision 2026-09-14). On each sweep, delete build-cache entries (`*-d`, `*-a` under `GOCACHE`) whose mtime is older than `max_age`; if the cache is still above `max_bytes`, delete oldest-mtime entries until under. Never delete `trim.txt`/`README`. Go tolerates missing entries (rebuild), so this is lossless. No `go clean -cache`.
6. Doctor and status print the host cache paths and sizes; that is the only observability added.

Depends on nothing. The invariant issue depends on this one.

## Acceptance

- Table-driven tests: worker env carries no `GOCACHE`/`GOMODCACHE`/`GOBIN`/`GOLANGCI_LINT_CACHE` override; writable roots include the resolved cache dirs under `workspace-write`; legacy roots removed at boot; trim deletes entries older than `max_age`, then evicts remaining entries oldest-first to `max_bytes`; recent entries are kept only while the cap allows. Never delete `trim.txt` or `README`.
- On mac-studio after deploy: `~/code/*/.detent/cache` absent; one host cache; `du` of `go env GOCACHE` stays under `max_bytes` across a day of fleet activity (measure, do not assume).

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

Workers inherit host toolchain caches while retaining per-attempt temporary directories. Workspace-write Codex sessions gain access to the resolved Go caches. Configuration no longer emits the removed cache strategy and warns when older files specify it; startup removes legacy project caches and reports reclaimed bytes.

Cache discovery runs once per process: daemon environment overrides, then `go env` from a neutral directory with `GOTOOLCHAIN=local`, then Go defaults. Discovery failure logs a warning and cannot prevent a Codex turn; the reaper skips unavailable caches.

The existing reaper expires build entries older than 48 hours, then evicts remaining entries oldest-first to enforce the configured cap (default 1000 GiB). The size cap takes precedence over age. Deletion requires a regular README or trim.txt marker and only considers full Go hash filenames directly under two-character hexadecimal shards. Other subtrees and metadata are preserved. Doctor and status report host cache paths and sizes.

Fixes #2680

Invariants touched: none. Operator approval under INV-11 is recorded on the issue. This removes the worker cache override and uses the existing reaper loop; the rework makes discovery optional and narrows the deletion predicate.

## UI Surface Contract

- [x] N/A: no dashboard layout or primary-screen changes. The issue authorizes doctor/status cache reporting.
- [x] No persistent top-of-screen messaging is added.
- [x] CLI behavior is covered by tests.

## Test Plan

- [x] Focused toolcache and Codex regression tests.
- [x] Table-driven environment, sandbox, policy, migration, age/cap eviction, metadata and unrelated-file preservation tests.
- [x] Missing/failing Go, daemon overrides, fallback defaults, neutral local-toolchain lookup, and process-cached discovery tests, including multiple provider turns without Go.
- [x] Final integrated `make check-fast` (includes generation, invariant checks, build, lint, vet, and full tests).
- [x] Current-head CI: [run 34924063816](https://github.com/digitaldrywood/detent/actions/runs/34924063816), successful PR placeholder checks on `f21bccb5`.
- [ ] Post-deploy mac-studio measurement across a day of fleet activity.

Concurrent builds can grow the cache between sweeps. Metadata and skipped subtrees are never deleted. No deployment or live-cache purge was performed. Race, coverage, and nilaway checks run in the merge queue under the assigned validation protocol.


corylanou:
Operator review: sending back to Rework. All three Codex findings were verified against the branch (`3eae31c`) and reproduced where applicable. The shared host cache design stays; only discovery, the trim predicate, and one doc sentence change.

## 1. Cache discovery must never block a Codex turn (P1, `internal/codex/sandbox.go`)

`AgentBackend.runTurn` calls `hostCacheWritableRoots` → `toolcache.Resolve` → `go env` for every non-restricted turn (default `thread_sandbox` is `workspace-write`), and any error aborts the turn before the provider is contacted. Reproduced two ways: `go` off `PATH` fails, and a `go.mod` in the daemon's cwd with an unavailable toolchain (`go 1.99`) makes `go env` exit 1 with "toolchain not available". The daemon here runs with cwd set to a Go module, so it inherits that module's toolchain selection.

Required change:
- Resolve once per process, not per turn.
- Order: `GOCACHE`/`GOMODCACHE` from the daemon environment when set → `go env -json` run from a neutral cwd (for example `os.TempDir()`) with `GOTOOLCHAIN=local` → Go's documented defaults (`os.UserCacheDir()/go-build`, `$GOPATH/pkg/mod` with `GOPATH` defaulting to `~/go`).
- If every step yields nothing, log one warning and proceed with no extra writable roots. `runTurn` must not return an error from discovery.
- Apply the same non-fatal resolution to the reaper's `TrimHostCache` in `internal/cli/runner.go` (a failed resolve skips the trim and logs; it already cannot abort the sweep).
- Tests: `go` missing, `go env` failing, env override honored, default fallback, and resolve called once across multiple turns.

## 2. Trim must only delete real Go build-cache entries (P1, `internal/toolcache/trim.go`)

The walk is recursive and matches any regular file whose name ends in `-a` or `-d`. A `GOCACHE` pointed at a broad directory would let the reaper delete unrelated files, oldest-first, unattended.

Required change:
- Candidates are only files directly under `root/<2 hex chars>/` whose name matches `^[0-9a-f]{64}-[ad]$`. No deeper recursion, and skip `fuzz/` and any other subtree.
- Refuse to trim a root that does not contain Go's own `README` or `trim.txt` marker; log and return without deleting.
- Tests: an unrelated `foo-a` at the root and in a nested dir survives; a root without markers is untouched; existing age/cap/metadata cases still pass.

## 3. Doc contradiction (P2, `docs/getting-started.md`)

The sentence "recent entries are protected even when their total exceeds this target" contradicts `toolcache.Trim`, `docs/config.md`, and this PR's own description. Replace with: the size cap takes precedence over age; recent entries are evicted oldest-first until the cache fits `max_bytes`.

None of this adds a mechanism: item 1 is a fallback chain for a lookup, item 2 narrows a predicate, item 3 is docs. Run `make check` and update the PR test plan.

Issue #2806, opened 2026-09-16
test(orchestrator): remove load-sensitive instant-failure truncation wait


corylanou:
Discovered during #2776 validation on base 751d7e3a. `make check-fast` failed after 18m27.412s solely in `TestRunParksIssueAfterRepeatedInstantBackendFailuresTruncatesConfiguredOutput` (1.03s), `internal/orchestrator/orchestrator_test.go:1585`: timed out waiting for orchestrator state. The test uses a fake connector and waits via `waitForState`, whose deadline is one second. The recorded event history shows four terminal_attempt_retry_demoted events between 10:13:42.629 and 10:13:43.434 on 2026-09-16 America/Chicago, with concurrent refresh stages around 200ms. No GitHub connector code participates in this fixture. Ten isolated repetitions passed in 0.572s; this suggests contention sensitivity but does not establish the full cause.

Acceptance: reproduce under concurrent package/test load and make the fixture synchronize on the behavior needed for truncation assertions rather than relying on a one-second scheduler deadline; retain all existing truncated body/comment assertions and production behavior. Validate focused tests/vet and the standard gate. No new operational mechanism.

#2776 is preserving its connector-only scope and will attempt the full gate with test parallelism reduced to one, without skipping tests. Original full gate log is retained in that attempt's TMPDIR as check-fast.log.

```detent-agent
schema: 1
effort: low
```

```detent-origin
origin_kind: worker
instance_identity: CorysMacStudio.localdomain/69005
source_ref: "5934"
fingerprint: orchestrator-instant-backend-truncation-test-one-second-state-timeout
```

corylanou:
## Detent backlog admission proposal

Evaluated at 2026-09-17T16:45:00Z. This is historical evaluation evidence; present readiness is rechecked before admission.

Proposal `admission-2b4222f0a425c449e3edbf92` recommends moving this issue to **Todo** and meets the configured automatic-admission threshold. Detent will move the issue to that state and record this proposal as the provenance.

Criteria section: **Admission Criteria**

- **Alignment** — Recorded make check-fast failure identifies the test, timeout, and event timestamps. The remedy removes a load-sensitive deadline in the fixture, retains assertions and production behavior, and adds no operational mechanism.
  - Criterion: “This includes
   flaky or timing-sensitive tests that fail CI or block a release
   cut — a red release gate is an operator-visible symptom.”
[...]

Issue #2754, opened 2026-09-15
test(github): validate multi-project refresh against the hourly API budget


corylanou:
## Evidence

One instance with five active projects, three consecutive hours of `github rest budget summary` lines:

| hour | requests | billable | 304 not-modified |
| --- | --- | --- | --- |
| 07 | 9823 | 4544 | 5279 |
| 08 | 10371 | 6680 | 3682 |
| 09 (partial) | 5271 | 3009 | 2262 |

The REST limit is 5000 per hour. Hour 08 burned 6680 billable requests, so the budget was exhausted well before the reset, and the instance spent the remainder of each hour below the 1000-request lookup floor. In the same window `github lookup backoff scheduled` fired 17, 17, 16, 16, 13 and 5 times for the six projects, and `/api/v1/state` showed five of them simultaneously carrying `github_lookup_backoff` with `refresh: degraded` and `last_error: tracker candidates data is stale`.

Consequence: with stale candidate data the scheduler sees almost nothing eligible. At 14:20Z one project had 13 candidates and 0 eligible; four others had 0 or 1 candidate against boards holding hundreds of Todo items. Only one project was dispatching at all, using 4 of its 5 permits, while every other project sat idle with work available. Throughput is limited by Detent's own REST accounting, not by capacity or by the agent provider.

Per-family `used` totals are dominated by list-shaped reads: repository issues 645889, issue reads 620330, issue comments 612795, label issues 608094, pull requests 504467, issue dependencies 389427. A third to a half of all requests come back 304 not-modified, which means the data was already current and the call was avoidable.

A restart makes it acute: the cold refresh of ten projects at 13:34Z consumed the budget within minutes and two dispatches then failed outright with `start state transition failed: github rate limited: status 429: GitHub GraphQL remaining budget is reserved for shared work`.

## Scope after operator-approved decomposition

Depends on: #2761
Depends on: #2762
Depends on: #2763
Depends on: #2776

This is the final workload-validation task, not an epic or a combined implementation PR. Implement batched hydration in #2761, revision-aware detail reuse in #2762, and replacement of global lookup-floor control with project refresh pacing in #2763. Native blocked-by relations mirror these declarations.

Correction from investigation: ETag/If-None-Match and zero-billable 304 accounting already work. Preserve them; there is no missing ETag accounting feature to implement.

## Expected

[...]

PR #2810, opened 2026-09-16
fix(github): reuse batched hydration for project refresh

### LINKED ISSUE #2776: fix(github): reuse batched hydration in normal project refresh
## Evidence

Discovered while validating #2754 after #2761/#2762/#2763 merged into main b6ddbaca. The optimized `ReadCandidates` path is used by admission, while normal orchestrator `fetchTickIssues` calls `FetchRefreshIssues`. `fetchProjectRefreshIssues` still calls `populateBlockerReasons`, `hydrateBlockedByRefs`, and `attachPullRequests` separately; it does not use candidateEvidence/revision-aware candidate PR hydration.

A local request-count fixture (five isolated connectors, three Human Review PR-linked issues/project, 120 refreshes at an injected 30-second cadence) measured:

| entry point | REST | billable REST | GraphQL | 304 |
| --- | ---: | ---: | ---: | ---: |
| ReadCandidates, board, no ETags | 30 | 30 | 1800 | 0 |
| ReadCandidates, labels, no ETags | 630 | 630 | 1800 | 0 |
| FetchRefreshIssues, no ETags | 4320 | 4320 | 600 | 0 |
| FetchRefreshIssues, stable ETags | 4319 | 419 | 600 | 3900 |

These are synthetic fixture counts, not live-hour results. The no-ETag arm is a stress case; the stable-ETag arm passes the billable threshold. Request totals varied by 1–2 between runs, so that timing variation needs resolving before freezing exact assertions. Existing conditional/free-304 behavior is correct. The local regression `TestCandidateHourlyWorkload/five_projects_hourly/refresh/legacy=false` fails the #2754 target of billable REST <2500 in the no-ETag arm.

## Scope

Consolidate normal project refresh hydration with the existing batched/revision-aware evidence path instead of leaving two independent implementations. Cover candidate and observed lanes and preserve native dependency authority, comment freshness, independent PR-state changes, pagination, fallback and ETag accounting. No new mechanism, config, CLI or dashboard surface.

## Acceptance

- Normal FetchRefreshIssues exercises the shared batch hydration rather than repeated per-issue REST hydration.
- Reproduce the workload above and demonstrate reduced requests, including when validators are absent or cannot yield 304.
- Preserve observable changes in comments, blockers and PR state after quiet refreshes.
- Use deterministic fixture clocks/counts and preserve free-304 accounting.
- Focused tests/vet and make check-fast pass.

This is a production follow-up to the operator-approved decomposition. #2754 remains the final workload-validation task and must not claim the admission-only counts prove the full refresh result.

```detent-agent
schema: 1
effort: low
```

```detent-origin
origin_kind: worker
instance_identity: CorysMacStudio.localdomain/12533
source_ref: "5911"
fingerprint: github-normal-refresh-bypasses-batched-candidate-hydration
```

corylanou:
## Summary

Normal project refresh now uses the same batched scheduler evidence and revision-aware PR hydration as admission. It preserves board pagination, native dependency authority, out-of-lane diagnostics, and fresh observed-lane fallback.

A deterministic five-project hour (three PR-linked issues/project, 120 ticks at 30 seconds) drops from 4,320 billable REST requests to 30 with no validators; GraphQL rises from 600 to 1,800. Candidate and observed lanes also use 30 REST with stable or changing validators. Unavailable GraphQL evidence retains legacy fallback and free-304 accounting; those cases can exceed the optimized bound. Full synthetic results and limitations are recorded in `docs/validation/github-project-refresh-hydration.md`; #2754 retains final workload validation.

Project scans share a single schema fallback. Only typed GraphQL undefined-field errors retry the reduced board query; rate limits, transport/context failures, and other GraphQL errors return directly. Both admission and normal refresh have request-count regression cases, including one HTTP request on 429.

`normalizeProjectItem` calls `buildIssue`, which can supply a non-empty `BlockerReason` from Workpad evidence before shared hydration. `applySchedulerEvidence` now matches legacy parsing: an empty parsed reason preserves the existing value; fresh reasons replace it. The existing native-authority step still recomputes reasons from current Workpad signals and removes stale dependency-only reasons. Other scheduler fields continue to refresh from complete evidence.

Fixes #2776

Invariants touched: none. This consolidates existing hydration; no new operational mechanism or configuration is added.

## UI Surface Contract

- [x] N/A: connector-only change; no UI surface changes.

## Test Plan

- [x] Reproduced the 4,320-request no-validator failure before implementation.
- [x] `go test ./internal/connector/github/... -count=1`
- [x] `go vet ./internal/connector/github/...`
- [x] Exact hourly/cold workload counts for candidate and observed refresh; HTTP counters match client total/billable/304 accounting.
- [x] Comment/dependency/PR changes after 120 quiet ticks across admission, candidate refresh, observed refresh, and overlapping lanes; pagination and native authority.
- [x] `make check-fast GO_TEST="env -u DETENT_API_TOKEN go test -count=1 -parallel=1"` (all tests retained; reduces contention). Default gate hit the unrelated one-second orchestrator fixture timeout tracked in #2806; 10 isolated repeats passed. Race/coverage/nilaway run in the merge queue.

Rework validation: `go test ./internal/connector/github/... -count=1` and focused vet passed; full `make check-fast GO_TEST="env -u DETENT_API_TOKEN go test -count=1 -parallel=1"` passed in 13m55.769s after fixing one staticcheck expression warning. Current-head CI run 35127990097 passed on d0015dc25825681aa78917e2a158de10d155cd19. All three operator review threads are resolved.


internal/connector/github/projects_v2.go
```diff
@@ -372,11 +384,23 @@ func (c *Connector) fetchProjectItemsScanWithLimit(
 			"first":     projectItemsPageSize,
 			"after":     after,
 		}, &response); err != nil {
-			return connector.IssueStateScan{}, fmt.Errorf("fetch github project items: %w", err)
+			if queryDocument != candidateProjectItemsQuery {
+				return connector.IssueStateScan{}, fmt.Errorf("fetch github project items: %w", err)
+			}
+			// Older schemas still provide board membership; shared hydration
```
corylanou: The schema fallback retries on any error class. A 429, a transient network failure, or a cancelled context on the candidate query issues a second full-cost GraphQL request per page, per project, exactly under rate-limit pressure. Gate the fallback on the field-unavailable classification only, and add a table case for a 429 that asserts a single request.

internal/connector/github/projects_v2.go
```diff
@@ -372,11 +384,23 @@ func (c *Connector) fetchProjectItemsScanWithLimit(
...
+			// falls back to REST when scheduler evidence is unavailable.
+			if fallbackErr := c.client.GraphQLWithType(ctx, queryType, observedStatusProjectItemsQuery, map[string]any{
+				"projectId": c.projectID, "first": projectItemsPageSize, "after": after,
+			}, &response); fallbackErr != nil {
+				return connector.IssueStateScan{}, fmt.Errorf("fetch github project items: %w", errors.Join(err, fallbackErr))
```
corylanou: candidate.go:239-246 keeps its own copy of this schema fallback. #2776 asked to stop carrying two independent implementations; hydration was consolidated, this was not. Share one helper between the two scans.

internal/connector/github/candidate_hydration.go
```diff
@@ -145,6 +145,20 @@ func (c *Connector) hydrateCandidateWithEvidence(ctx context.Context, issue conn
 		err := c.hydrateCandidateIssues(ctx, issues, states)
 		return issues[0], err
 	}
+	issue, err := c.applySchedulerEvidence(issue, node)
+	if err != nil {
+		return issue, err
```
corylanou, line 150: Confirm: applySchedulerEvidence overwrites Description, UpdatedAt, ModelOverride, Comments, WorkpadSignal and BlockerReason for every evidence-backed issue, where the replaced populateBlockerReasons (issue_read.go:1367-1391) touched only Blocked-state issues and never cleared a non-empty BlockerReason. If no refresh caller sets BlockerReason earlier, say so in the PR body and resolve; otherwise preserve the non-empty value.

corylanou (commented):
Operator hand review (no bot review capacity). The main change is right: fetchProjectRefreshIssues scans with candidateProjectItemsQuery, builds per-page evidence, and shares applySchedulerEvidence and hydratePullRequestWithEvidence with admission, with legacy REST as the per-issue fallback; TestProjectRefreshHourlyWorkload and the extended candidate-evidence test cover the workload. Two threads to fix before merge and one to confirm. No new mechanism, config key, or reason code; docs/invariants.md needs no change (transport moved, rule unchanged).

internal/connector/github/projects_v2.go
corylanou: Fixed in d0015dc2. The shared page helper retries only typed GraphQL undefined-field errors; generic/mixed GraphQL, 429/rate-limit, transport, cancellation and deadline errors return directly. Both entry points have a request-count table, including exactly one HTTP request on 429. The original client backoff already suppressed the second 429 HTTP request; the new helper also removes that fallback attempt. Focused tests/vet and full check-fast pass.

internal/connector/github/projects_v2.go
corylanou: Fixed in d0015dc2. Both readProjectCandidates and scanProjectItems call queryProjectItemsPage with the original query type, pagination variables and output. The reduced-query fallback and error classification now live in that one helper. Admission and normal refresh regressions and workload tests pass.

internal/connector/github/candidate_hydration.go
corylanou: Confirmed that normalizeProjectItem calls buildIssue, which can set BlockerReason from Workpad evidence before hydration. Fixed in d0015dc2 to match the legacy parsing behavior: an empty parsed reason preserves the existing value; a fresh non-empty reason updates it. The existing native-authority step still recomputes reasons from current Workpad signals and removes stale dependency-only reasons. Added retain/update/populate/empty regression cases and documented this explicitly in the PR body. Full connector tests and check-fast pass.

PR #2819, opened 2026-09-16
fix(github): keep project refresh within GraphQL budget

### LINKED ISSUE #2818: fix(github): whole-board refresh carries the enriched candidate payload and exhausts the GraphQL hour on large ProjectV2 boards
## Evidence

Regression from #2776 (v0.114.20, this instance since 18:12Z on 2026-09-16). A ProjectV2 project with 152 candidates and about 1,500 observed-lane items now exhausts the hourly GraphQL budget:

- `github graphql budget summary` cycles (local time): 13:20 cost 747 (observed_status 40 queries, 640 pts); 13:43 cost 787 (observed_status 20 queries, 600 pts); 14:10 cost 752 (observed_status 40, 640); 14:28 cost 520 (observed_status 15, 450); 14:33:20 cost 526; 14:33:49 cost 904 (a candidate_issues cycle). 8,505 points in 105 minutes on one instance; `graphql_remaining` fell from 3554 at 18:18Z to 976 at 19:40Z.
- Effects: four consecutive candidate scans for that project failed with `fetch github project items: github rate limited: status 429: GitHub GraphQL rate-limit response is in backoff` (19:02Z, 19:28Z, 19:30Z, 19:33Z), then `GraphQL remaining budget is reserved for shared work` (`client.go:722-732`); candidates source `failure_streak 4`; no dispatch decision since 19:03Z; five issues an external contributor tagged for this instance never appear. The second instance (65 candidates) received a GraphQL 403 on the same scan at 19:36Z and dropped 2191 to 1348 points in 13 minutes.
- Before #2776 the same project starved on the REST dependency fanout cap (#2776's evidence). The starvation moved from REST to GraphQL.

## Cause (verified at v0.114.21, Codex-challenged)

1. Normal project refresh runs `candidateProjectItemsQuery` for the entire board before lane selection (`internal/connector/github/issue_read.go:531`), under the `observed_status` accounting category. Every one of the ~1,500 observed-lane items therefore carries the full scheduler payload, not a status read.
2. That payload is expensive by connection count (`internal/connector/github/candidate_hydration.go:12` replaces `comments { totalCount }` with `comments(first: 100)` including bodies, adds `blockedBy(first: 20)` with each blocker's `labels(first: 100)`, and keeps `closedByPullRequestsReferences(first: 5)` with `commits(last: 1)`): up to 2,000 blocker-label connections and 500 commit connections per 100-item page. GitHub prices by connections, and the `rateLimit.cost` field is the actual measurement.
3. Cost attribution is partly blind: the generated candidate PR queries (`candidate_pull_requests.go:33`, `contexts(first:100)` with `annotations(first:100)`) omit `rateLimit.cost`, so the client substitutes header deltas (`client.go:1431`) and can attribute other requests' consumption to one query.
4. A page error inside `fetchProjectItemsScanWithLimit` (`projects_v2.go:394`) returns an empty scan and the next refresh restarts from page one, so every failed scan re-spends the pages it already fetched. The admission candidate reader already carries a cursor and returns its position on error (`candidate.go:222`); the two paths disagree.

## Fix (query-shape and consumer consolidation, no new mechanism, timer, or config)

- Whole-board enumeration must not carry the scheduler payload. Enumerate the board with the thin fields the normalization actually consumes (identity, repository, title, updatedAt, state, labels, priority, `statusValue`/`priorityValue` via `fieldValueByName`), then apply the enriched candidate hydration only to items in active lanes, reusing the batched hydration #2776 introduced. Measure with `rateLimit.cost`; do not assume the saving.
- Trim the candidate page: drop the `closedByPullRequestsReferences` preview (the successful PR-evidence path fetches associations and authoritative PR status separately) and bound `blockedBy` label fetching to what dependency evaluation needs; keep comments and native blockers complete, since `applySchedulerEvidence` consumes them in place of REST hydration.
- Add `rateLimit { cost remaining }` to every generated candidate and PR query so accounting is per request, never header deltas.
- Make the combined refresh reuse the existing candidate cursor contract so a failed page does not force a restart from page one on the next refresh. This consolidates two scan paths; it must not become a retry loop or a partial-snapshot publisher.

## Acceptance

- `TestCandidateColdRequestCounts` (`candidate_batch_test.go:15`) asserts the query shapes and sums `data.rateLimit.cost` from fixtures grounded in measured costs.
- `TestProjectRefreshHourlyWorkload` (`refresh_workload_test.go:21`) adds a 1,500-item board with 152 candidates at four refreshes per hour and asserts point totals alongside request totals: under 1,000 points per hour for that board.
- `TestCandidateBatchedPaginationAndAuthority`, `TestCandidateEvidenceIndependentConnections`, `TestCandidateBoardBatchedCursor` cover completeness, independent cursors, and failure position for the combined refresh.
- On this instance after deploy: the large project's refresh completes every cycle with `graphql_remaining` staying above the reserve across a full hour, and its tagged Todo issues receive scheduler decisions.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- Enumerate ProjectV2 boards with thin lane metadata, then hydrate scheduler evidence only for candidate and relevant observed lanes. Remove the redundant candidate PR preview and reduce blocker label pages while retaining completeness fallback.
- Report request-level GraphQL costs for generated scheduler/PR queries. Measured response fixtures put a 1,500-item board with 152 candidates at 252 points/hour across four refreshes (68 GraphQL and 4 REST requests).
- Reuse the candidate cursor contract in the board scanner so a failed page resumes without publishing a partial snapshot. Preserve independent connection cursors, REST body fallback, and native human-prerequisite evidence.

Fixes #2818

Invariants touched: [INV-3](docs/invariants.md#inv-3--mechanism-moratorium). This consolidates the existing scan and hydration consumers; no new retry loop, timer, configuration, or lane writer.

## UI Surface Contract

- [x] N/A: no UI changes.
- [x] This PR adds no persistent top-of-screen messaging.
- [x] No visual changes to primary operator screens; browser verification is not applicable.

## Test Plan

- [x] Reproduced the expensive whole-board query with a failing regression before implementation.
- [x] Focused candidate, refresh, pagination, schema fallback, and PR-authority tests; connector `go vet`.
- [x] `make generate` (no generated differences).
- [x] `make check-fast` passed in 378s (configured worker gate; race/coverage/nilaway belong to the merge queue).
- [x] Bounded live query measurements using `data.rateLimit.cost`: thin 100-item page=2, enriched 10-item admission page=2, scheduler batches of 100/52/3=22/11/1 points.
- [ ] Post-deployment: observe a full hour with budget above reserve and scheduler decisions for tagged Todo issues. No live process changes were made during implementation.


internal/connector/github/issue_read.go
```diff
@@ -527,39 +527,70 @@ func (c *Connector) fetchProjectRefreshIssues(
...
-			if _, wanted := wantedStates[normalizeStateName(state)]; wanted {
-				node.CandidateState = state
-			}
-			nodes = append(nodes, node)
-		}
-		for id, node := range c.candidateEvidence(ctx, nodes, false) {
-			evidence[id] = node
+	schedulerStates := normalizedStateSet(candidateStates)
+	for _, state := range observedStates {
+		if stateInList(state, c.terminalStates) || normalizeStateName(state) == normalizeStateName("Backlog") || normalizeStateName(state) == normalizeStateName("Done") || normalizeStateName(state) == normalizeStateName("Cancelled") {
```
corylanou, line 532: Lane names are hardcoded ("Backlog", "Done", "Cancelled") instead of the project's configured observed and terminal states. A project whose terminal lane is named differently and is absent from terminal_states gets full enrichment of that whole lane (the cost regression returns), and a project that uses "Done" as an active lane silently loses scheduler evidence. Derive the thin-versus-enriched split from the configured active_states, observed_states and terminal_states, and add a table case with non-default lane names.

internal/connector/github/issue_read.go
```diff
@@ -527,39 +527,70 @@ func (c *Connector) fetchProjectRefreshIssues(
...
+			node.Labels.Nodes = append(node.Labels.Nodes, label{Name: name})
+		}
+		nodes = append(nodes, node)
+	}
+	for start := 0; start < len(nodes); start += projectItemsPageSize {
+		for id, node := range c.candidateEvidence(ctx, nodes[start:min(start+projectItemsPageSize, len(nodes))], true) {
+			evidence[id] = node
+		}
+	}
 	var fallback []connector.Issue
```
corylanou, line 568: When candidateEvidence yields nothing (blockedBy unsupported, or a GraphQL backoff, which is exactly when this project is under pressure), every scheduler-lane issue falls back to a per-issue REST GET because the thin query no longer returns body. On the 152-candidate board that is 152 REST calls per refresh, about 608 per hour: the REST fanout #2776 removed comes back on the failure path. Bound this: fall back only for issues whose body is actually needed this cycle, or carry the body in the thin page for scheduler-lane items only, and extend TestRefreshThinBodyFallback to 152 candidates asserting the REST count stays under the per-family cap.

internal/connector/github/projects_v2.go
```diff
@@ -382,32 +388,51 @@ func projectSchedulerFieldsUnavailable(err error) bool {
 	return true
 }
 
+// An unfinished refresh uses the admission cursor contract and retains its
+// private accumulator. Callers publish only the completed snapshot.
+type projectItemsScanProgress struct {
```
corylanou, line 393: Resume reuses the previous cycle's cursor and captured revision with no board-revision invalidation. ProjectV2 POSITION pages shift between refreshes, so a resumed scan can omit items and still publish a complete LaneSignalCandidates snapshot. Either invalidate the retained progress when the board revision or item count changes, or mark the resumed scan as complete only when the final page's pageInfo confirms it; add a test where the board changes between the failed and resumed refresh and assert no item is dropped.

corylanou (commented):
Operator hand review (no bot review capacity). The shape is right: thin whole-board enumeration, enrichment only for scheduler-lane items batched through candidateEvidence, PR preview dropped, blockedBy labels bounded, rateLimit cost on every generated query, cursor reuse without a retry loop, invariants updated, and the workload test measures 252 points/hour for the 1,500-item board. Package tests pass on the branch. Three threads to fix before merge; the second one reintroduces the exact starvation #2776 removed.

internal/connector/github/issue_read.go
corylanou: Fixed in 100adeb4. Enrichment now uses requested candidates and configured active/observed states, with configured terminal states excluded. Unconfigured observed lanes remain thin. TestRefreshConfiguredSchedulerStates covers custom active/observed/terminal names, an unconfigured terminal-like lane, and Done/Backlog repurposed as scheduler lanes.

internal/connector/github/issue_read.go
corylanou: Fixed in 100adeb4 by retaining scalar bodies and comment counts in the board query and deleting per-issue REST body GETs. ProjectV2 cannot condition content fields on lane values, so the scalar body is retained for all board items; it adds no connections. A live 100-item request including body and project updatedAt still measured 2 points via data.rateLimit.cost. TestRefreshThinBodyFallback now covers 1 and 152 candidates for unsupported schema, incomplete evidence, and GraphQL backoff, asserting zero body GETs and at most 40 REST calls under the existing cap. If complete dependency hydration exceeds that cap, it returns the existing deferral without publishing a partial snapshot.

internal/connector/github/projects_v2.go
corylanou: Fixed in 100adeb4. Before resuming, the scanner verifies project updatedAt and item count (a measured 1-point metadata read), plus local cache revision. Changed or missing revisions discard retained progress and enumerate from page one; changes detected during page enumeration discard the scan and return ErrProjectItemsTruncated without publication. Unchanged scans retain the failed-page cursor. TestRefreshBoardChangedBetweenAttempts covers reordered pages, count changes, missing revision, and changes during resume; unchanged-page and cancellation regressions also pass. No retry loop is added.

Issue #2825, opened 2026-09-16
fix(github): resumed board scans fail on any mid-scan change and the default config re-enriches the Backlog lane


corylanou:
## Evidence

PR #2823 (fix for #2820) merged on 2026-09-16 at ~21:5xZ before its hand review finished (no automated reviewer; see #2822 for the queue behavior). The review found four defects that ship in the next release unless fixed first. Rollout of the release containing #2823 is held by the operator until this lands.

## Findings to fix

1. **Mid-scan invalidation can starve a busy board forever.** `internal/connector/github/projects_v2.go:473-479` rechecks the project's `updatedAt` between pages and fails the whole scan with "project changed during scan" on any change. Detent itself writes lane state on every tick, so a 15-page board that changes once per minute never publishes a snapshot. `TestRefreshBoardChangedBetweenAttempts` exercises a single shift, never a steady stream of changes. Invalidate retained resume progress only; a scan that started fresh must complete on the pages it fetched (a board that changed during the scan is stale by at most one refresh, which is the existing contract), or confirm completion by the final page's `pageInfo` as #2820 asked.
2. **Backlog enrichment regression under the default config.** `internal/connector/github/issue_read.go:530-539` derives the enriched set from terminal plus active plus observed states; with the default `observed_states` [Backlog, Human Review, Blocked] the whole Backlog lane is enriched again (70 items on the largest projects), which is the overspend #2818 removed. Enrich active lanes only (plus Human Review and Blocked if the routing paths need their evidence, with the reason stated in the PR); Backlog is never a dispatch candidate. Pin it: the workload fixture in `TestProjectRefreshHourlyWorkload` must include a Backlog lane and assert its items receive no enrichment.
3. **`items(first: 0)` in `refreshProjectRevisionQuery` is unvalidated against the live schema.** Only fixtures exercise it; if GitHub rejects the argument, every resumed scan fails. Either use a documented shape (drop the connection and read `updatedAt` only) or validate it against the recorded live schema in the connector's contract tests.
4. **The preflight revision query is not in the hourly budget.** One extra GraphQL round trip per resumed refresh is absent from `TestProjectRefreshHourlyWorkload`; add it to the fixture and the asserted totals.

No new mechanism, config key, timer, retry loop, or reason code. `docs/invariants.md:587-600` updated if the cursor contract changes.

## Acceptance

[...]

PR #2832, opened 2026-09-16
fix(github): honor configured refresh routing and completeness

### LINKED ISSUE #2830: fix(github): thin refresh still keys enrichment on literal lane names and skips the completeness check
## Evidence

PR #2827 (fix for #2825) fixes the mid-scan starvation and the Backlog enrichment for the default configuration, but three gaps remain on main (hand review, 2026-09-16 23:0xZ):

1. `internal/connector/github/issue_read.go:534` excludes Backlog by literal name before the candidate/active check. A project with Backlog in `active_states` (or Backlog passed as a candidate state, `tick.go:764` returns `cfg.ActiveStates`) gets an empty scheduler set and dispatches from Backlog with no PR, blocked-by, or Workpad evidence. The PR changed `candidate_batch_test.go` refresh mode from Backlog to Todo and flipped the "backlog repurposed" expectation to `enrich=false`, encoding the regression.
2. `issue_read.go:537` hardcodes ["Human Review", "Blocked"] for observed-lane enrichment; a board whose review lane has another name in `observed_states` loses routing evidence. `refresh_query_test.go` "custom observed" (observed_states ["Review"]) was flipped from `enrich=true` to `false`.
3. The thin refresh no longer calls `validateProjectItemsComplete` (`projects_v2.go:509-514`), so an enumeration truncated by GitHub (hasNextPage=false after page 3 of 15) publishes a partial snapshot as complete and every unseen item reads as absent from lane signals.

## Fix (no new mechanism, config key, timer, or reason code)

- Derive both splits from configuration: the enriched set is the configured active states plus the configured observed states that the routing paths consume (name them from the auto-promote and blocked-recovery readers), never a literal lane name; observed-only states are thin. Restore the flipped test expectations as the correct behavior.
- Keep a completeness check for the thin enumeration: compare the enumerated count against the project's item totalCount (already fetched by the revision preflight or the first page) and mark the snapshot incomplete on mismatch, using the existing incomplete-snapshot path.

## Acceptance

Table-driven, standard-library tests: Backlog in active_states enriches; observed lane under a custom name that the routing readers consume enriches; observed-only lanes stay thin; a truncated enumeration is not published as complete. `go test ./internal/connector/github/` passes; docs/invariants.md updated if the cursor or completeness text changes.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- Enrich active/candidate lanes regardless of their names, including Backlog. Pass observed routing states from existing auto-promote and recovery readers; leave metadata-only lanes thin.
- Restore the ProjectV2 completeness check against the first-page item count. Truncated enumerations publish no snapshot; later count growth and timestamp changes allow the scan to complete.
- Preserve the observed PR fallback status policy to avoid extra REST fanout for active/candidate lanes.

Fixes #2830

Invariants touched: [INV-3](docs/invariants.md#inv-3--mechanism-moratorium). Consolidates enrichment selection and restores the existing completeness check; no new operational mechanism or configuration key.

## UI Surface Contract

- [x] N/A: no UI changes.

## Test Plan

- [x] Reproduced active Backlog, custom Review, and truncated enumeration failures before implementation.
- [x] Reproduced both operator rework findings before fixing: count growth rejected complete scans; observed PR fallback doubled REST calls.
- [x] Standard-library table tests cover configured routing, thin metadata-only lanes, repeated busy-board scans, truncation/retry, and REST request counts.
- [x] Connector package tests and vet; initial implementation also passed orchestrator tests and vet.
- [x] Final rework `make check-fast` passed in 482 seconds (race/coverage/nilaway run in merge queue).


internal/connector/github/projects_v2.go
```diff
@@ -506,12 +506,10 @@ func (c *Connector) scanProjectItems(
 		}
 
 		if !response.Node.Items.PageInfo.HasNextPage {
-			// A thin refresh completes at the final page, even if concurrent board
-			// edits changed totalCount. Other scanners retain their count contract.
-			if queryDocument != thinRefreshProjectItemsQuery {
-				if err := c.validateProjectItemsComplete(ctx, scan.ItemsFetched, scan.TotalItems); err != nil {
-					return connector.IssueStateScan{}, err
-				}
+			if err := c.validateProjectItemsComplete(ctx, scan.ItemsFetched, scan.TotalItems); err != nil {
```
corylanou, line 515: This reintroduces the #2825 starvation. scan.TotalItems is kept as the max across pages, so any mid-scan change to the board's totalCount (one issue added or removed while paginating a 15-page board) yields fetched < total, ErrProjectItemsTruncated, and the progress reset discards the cursor; the next tick re-enumerates and fails the same way. The PR encodes it: TestRefreshBusyBoardCompletes "count" was flipped to assert three consecutive failures with zero statuses. Use the totalCount from the preflight or the first page and tolerate growth; a board that changed during the scan is stale by one refresh, which is the existing contract. Restore the test to assert the scan completes.

internal/connector/github/issue_read.go
```diff
@@ -664,7 +676,7 @@ func (c *Connector) hydrateRefreshPullRequests(ctx context.Context, issues []con
 	if candidates {
 		err = c.attachPullRequests(ctx, fallback)
 	} else {
-		err = c.attachStatePullRequests(ctx, fallback, false)
+		err = c.attachFreshPullRequests(ctx, fallback)
```
corylanou: attachFreshPullRequests replaces attachStatePullRequests(..., false) and bypasses pullRequestStatusPolicy, so every candidate and active-lane issue now takes the REST PR fanout (candidate_batch_test.go wantREST went from 1 to 2). #2830 did not ask for this and it raises REST cost on every refresh. Keep the status policy (Human Review, Merging, Blocked) and restore wantREST.

corylanou (commented):
Operator hand review (no bot review capacity). The configuration-derived enrichment is right (routing states from the auto-promote, plan-stop, dependency-unblock and blocked-recovery readers; Backlog thin under the default config; flipped expectations restored; invariants updated). Two threads block the merge; I dequeued the PR and moved the card to Rework.

internal/connector/github/projects_v2.go
corylanou: Reproduced with the restored success assertion: 15 fetched items were rejected against a later-page total of 29. Thin refresh now retains the first-page totalCount as its completeness baseline; later growth cannot raise it. The count fixture starts each scan at 15, grows on later pages, and requires three consecutive scans to complete. The separate truncation/retry table still rejects 3 fetched items against an initial count of 15. Focused regressions, connector tests, and vet pass; final check-fast is running.

internal/connector/github/issue_read.go
corylanou: Restored attachStatePullRequests(ctx, fallback, false), preserving pullRequestStatusPolicy for fresh observed reads. Restored wantREST=1 in the candidate/observed Backlog fixture; it failed with REST=2 before the fix and passes now. Connector tests and vet pass; final check-fast is running.

PR #2851, opened 2026-09-17
fix(dispatch): recover GitHub monitor probes after throttling

### LINKED ISSUE #2846: fix(dispatch): the worker GitHub budget monitor stops a whole project on one rate-limited probe and never re-probes
## Evidence

Instance running five projects and Codex workers on one GitHub account, v0.114.25 then v0.114.26, 2026-09-17:

- 01:45:52Z a worker session for a ProjectV2 card failed with `worker github REST budget monitor failed: operation=credential_classification_worker consumer=shared_pool credential_identity=github-rest:9c4cd46adfa9: github authenticated user graphql probe returned HTTP 403`. The 403 was GitHub's secondary rate limit, which the account was tripping every few minutes at the time; the credential itself was fine (it kept serving the orchestrator).
- 01:46:29Z `worker GitHub REST budget monitor unavailable`, `next_probe_at 01:50:52Z`, `probe_attempts 0`; the attempt was completed as terminal `capacity`, error class `worker_github_budget_monitor_unavailable`, phase `waiting`, with `worker_github_monitor_wait` metadata.
- From 01:49:45Z every scheduler decision for the project's 47 cards, including the card that carries the monitor's retry, was `worker_github_budget_monitor_unavailable`. No canary probe was ever logged. The service restarted at 01:57Z and the hold was restored from the attempt metadata. Zero sessions ran for over 40 minutes on a board with five slots and 47 admitted cards.
- The operator cleared it by hand: the attempt row's error class was changed so the restore path ignores it, and the service was restarted at idle.

## Cause (verified on origin/main; Codex challenge in progress, corrections will be appended)

1. `internal/orchestrator/worker_github_monitor.go` `workerGitHubMonitorBlocks` (~:232): for an issue without a GitHubMonitor retry, any registered condition whose `ProbeIssueID` is not this issue blocks dispatch. Only a retry with `retry.GitHubMonitor` for that credential can reserve the probe (`reserveWorkerGitHubMonitorProbe` ~:249). If the probe-carrying retry is missing, replaced, or is itself skipped earlier in `dispatchIssue` (`dispatch.go` ~:583 order), no issue can ever probe and the project is blocked with no expiry.
2. `restoreWorkerGitHubMonitorWait` (~:405) re-registers the condition from attempt metadata after a restart, so the hold survives restarts; the carrier retry it recreates did not lead to a probe (the carrier card was skipped with the monitor reason at 02:03Z and 02:12Z).
3. A single HTTP 403 on the credential-classification probe during a secondary-limit cooldown is treated as a credential budget-monitor failure and converted into a project-wide dispatch stop. Rate-limited or in-cooldown probe results are not distinguished from an actual credential failure (compare the client's secondary handling with Retry-After in `client.go`).

## Fix (consolidation; no new brake, timer, config key, or reason code)

- A rate-limited or in-cooldown classification probe must not register the monitor; it waits on the existing secondary cooldown like every other GraphQL call.
- The monitor must never block the issue that carries its retry, and a condition with no carrier must expire at `NextProbeAt` rather than hold forever (consolidate with the existing backend-capacity probe lifecycle).
- Restoring from attempt metadata must recreate the carrier retry so the probe actually runs at `NextProbeAt`.

## Acceptance

Extend `worker_github_monitor_test.go` and `dispatch_test.go`, table-driven and standard library: a secondary-limit 403 on classification registers no monitor; a registered monitor's carrier is dispatched as the probe at `NextProbeAt`; a monitor without a carrier expires; after a restart the restored condition probes at `NextProbeAt` and clears on success. `docs/invariants.md` updated where the monitor or backend-capacity probe text describes this.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- Worker credential classification shares the connector's secondary cooldown and respects Retry-After. Throttling waits and retries without registering a monitor failure; cancellation remains cancellation.
- Monitor carriers survive missing candidate batches, including after durable recovery. Probe reservation happens only after eligibility and worker selection succeed.
- Idle holds expire at the existing probe deadline even when their carrier cannot dispatch. Eligibility is read-only; retained attempt history preserves escalating backoff on another failure. Successful carrier probes clear the condition.

Fixes #2846

Invariants touched: INV-2 ([same-PR invariant update](https://github.com/digitaldrywood/detent/blob/21e88c3b46c9a7658e71cdd8004056c517bf9c23/docs/invariants.md#inv-2--instance-owned-infrastructure-failures)). Consolidates classification onto the shared cooldown and removes carrier-dependent indefinite holds. No new mechanism, configuration, or reason code.

## UI Surface Contract

- [x] N/A: no UI changes.

## Test Plan

- [x] Reproduced throttling, lost carrier, and orphan-hold failures before the initial fix.
- [x] Reproduced rework findings: empty candidate batches discarded live/restored retries, eligibility discarded backoff history, and ineligible carriers renewed expired holds.
- [x] Table-driven regressions cover cooldowns, cancellation, durable recovery through successful dispatch, missing candidate batches, retained backoff, and inactive/terminal/dependency-blocked carriers.
- [x] Affected orchestrator tests and vet pass.
- [x] Final configured gate: make check-fast passed in 469 seconds. Earlier runs caught a corrected test lint issue and an instance cache failure (follow-up #2853); the affected cache-dependent packages passed before the successful full rerun.


internal/orchestrator/worker_github_monitor.go
```diff
@@ -230,19 +230,41 @@ func workerGitHubMonitorBlocks(state *State, issueID string, retry Retry, now ti
...
+			for _, queued := range state.Retry {
+				if queued.GitHubMonitor && strings.TrimSpace(queued.GitHubCredential) == key {
+					hasCarrier = true
+					break
+				}
+			}
+			// A lost or replaced retry must not leave a permanent project hold.
+			// Keep the existing probe deadline as the orphan condition's expiry.
+			if !hasCarrier {
+				delete(state.GitHubMonitors, key)
```
corylanou: Item 3 is missing: releaseMissingDueRetries (dispatch_planner.go:594) is unchanged and its only preserve hook (dispatch.go:307) covers Merging-lane retries, so a monitor carrier absent from one candidate batch is still released and the condition is orphaned; that is the path that stranded #2171 on 2026-09-17. Preserve monitor-carrier retries there (or expire the condition when its carrier is released) with a dispatch regression for a missing candidate batch. Also: expiry deletes the condition, so ProbeAttempts resets to 0 and re-registration always uses backendCapacityProbeDelayForAttempt(0); a genuinely bad credential with no carrier cycles fail, expire, dispatch, fail at the minimum delay forever. Keep the attempt count (or the escalated delay) across expiry.

internal/orchestrator/worker_github_monitor.go
```diff
@@ -230,19 +230,41 @@ func workerGitHubMonitorBlocks(state *State, issueID string, retry Retry, now ti
...
-		if condition.ProbeIssueID != issueID {
-			return true
+	blocked := false
+	for key, condition := range state.GitHubMonitors {
+		if condition.ProbeIssueID == issueID {
+			continue
+		}
+		if condition.ProbeIssueID == "" && !now.Before(condition.NextProbeAt) {
+			hasCarrier := false
+			for _, queued := range state.Retry {
```
corylanou: Two more: workerGitHubMonitorBlocks now mutates state.GitHubMonitors from read-shaped paths (dispatch_planner.go:751 and the API-receipt path in work_attempt_retry_intent.go:261); confirm both run under the same state lock as the dispatch loop or move the expiry to the tick. And the carrier scan accepts any retry with the credential regardless of whether that issue is still dispatchable, so a permanently blocked carrier suppresses expiry, the original hold narrowed; require the carrier to be dispatchable or expire anyway.

corylanou (commented):
Operator hand review (no bot review capacity). Items 1 and 2 are done well (the classification probe now goes through the connector client and honors Retry-After; orphaned conditions expire at NextProbeAt). Item 3 is not done and two of the new behaviors need adjustment; dequeued and moved to Rework.

internal/orchestrator/worker_github_monitor.go
corylanou: Fixed in 21e88c3b4. The existing preserveMissingDueRetry hook now retains monitor carriers across missing candidate batches. The live/restored dispatch regression runs an empty batch before the successful probe and recovery. Idle expiry no longer deletes the condition: it retains attempt history, and a subsequent failure advances backoff. Table-driven tests cover attempts 0, 2, and 5. Affected tests/vet and make check-fast pass (469s).

internal/orchestrator/worker_github_monitor.go
corylanou: Fixed in 21e88c3b4 by removing mutation and the carrier scan from the predicate. Idle holds expire at NextProbeAt regardless of carrier dispatchability while retaining backoff history. Probe reservation now follows eligibility and worker selection, so a rejected carrier cannot renew the hold. Regressions cover Backlog, Done, and an active carrier blocked by a dependency, plus unchanged monitor history after eligibility reads. Affected tests/vet and make check-fast pass (469s).

PR #2863, opened 2026-09-17
fix(github): eliminate ProjectV2 dispatch hydration fanout

### LINKED ISSUE #2859: fix(github): ProjectV2 refresh leaves Fields empty so dispatch re-reads every candidate over REST before authorization
## Evidence

Recorded on a ProjectV2 board with ~150 open candidates (one instance, v0.114.30, 2026-09-17 05:07Z to 05:37Z):

- Every refresh logs `hydrate dispatch issue failed ... github rest fanout deferred: endpoint family issue reads reached request cap 500` for the tail of the candidate list: 57 candidates at the 05:07Z tick, 132 at the 05:36Z tick. The `scheduler_decisions` rows for them carry reason `hydrate_failed` with no identifier.
- The `reconciliation` stage of that project's refresh took 2m39s (05:08Z) and 3m51s (05:32Z); the whole refresh took 6 to 8 minutes, and the other project on the same host logged `orchestrator tick loop frozen` twice (missed_intervals 2) while it ran.
- The cap consumed by dispatch hydration then starves live sessions: `session progress heartbeat failed: fetch github issue comments: github rest fanout deferred: endpoint family issue comments reached request cap 500` three times for one running session in 10 minutes.
- REST remaining was 3214 to 4004 of 5000 at the time; the limiter is the per-refresh fanout cap, not GitHub.

## Cause (verified by reading `origin/main` 6641d568)

`fetchProjectRefreshIssues` uses `thinRefreshProjectItemsQuery` (`internal/connector/github/issue_read.go:558`), which reads Status and Priority through named aliases, while the converter builds `Issue.Fields` only from `item.FieldValues` (`internal/connector/github/projects_v2.go:592`). Every ProjectV2 refresh candidate therefore reaches dispatch with an empty `Fields` map.

`hydrateDispatchIssue` (`internal/orchestrator/dispatch.go:355`) treats `len(issue.Fields) == 0` as "not hydrated" and calls `FetchIssueStatesByIDs` for that one issue, which performs a REST `GET /repos/{owner}/{repo}/issues/{number}` (`issue_read.go:1432`) plus the project-field read, and `hydrateDispatchDependencies` follows with dependency reads. The planner runs this hook (`internal/orchestrator/dispatch_planner.go:191`) before `dispatchAction` (line 209), where the authorization selector (`dispatchSkipAuthorizationSelector`, line ~782) declines candidates that this instance is not authorized for. On this board 89 of the candidates are declined by the selector every tick, after paying the REST read.

Net effect: N candidates cost N per-issue REST GETs per refresh, before any authorization or eligibility decision, on every ProjectV2 instance in the fleet (`issue_read.go:503` routes all ProjectV2 refreshes through this path). The batched candidate hydration added by #2776/#2810/#2849 does not cover project field values, so it does not prevent this.

## Fix

Consolidate project-field retrieval into the existing paginated refresh query: include the supported `fieldValues` selection (single-select, text, number, field name, `updatedAt`; the selection already exists at `projects_v2.go:97`, conversion at `projects_v2.go:1081` and `issue_read.go:1700`) in each refresh page so `Issue.Fields` and `FieldUpdatedAt` are populated by the scan and the existing `len(Fields) > 0` short-circuit in `hydrateDispatchIssue` holds for board candidates. Do not populate only the Status/Priority aliases: custom-field selectors (`internal/selector/match.go:315`) and the artifact status field and timestamp (`autopromote.go:459`, `dispatch_planner.go:870`) need the full collection. Keep the scheduler-evidence flow of the thin query intact.

Additionally, evaluate the label-only part of the authorization selector before the hydrate hook so declined candidates never reach hydration; keep field-dependent authorization and full eligibility after hydration. Reuse `dispatchSkipAuthorizationSelector`.

No new mechanism, cache, configuration key, or reason code. This removes work; it does not add a guard.

## Acceptance criteria

- A regression test drives a real GitHub connector against a fake HTTP server returning 150 ProjectV2 candidates across two pages, passes the actual `FetchRefreshIssues` output through the production dispatch hydration hooks, and asserts zero `GET /repos/{owner}/{repo}/issues/{number}` requests during dispatch across at least two refresh/dispatch cycles, with all 150 candidates reaching the hydrate hook (no vacuous pass from exhausted capacity).
- Custom-field selector matching and non-matching, artifact wait and timestamp decisions before and after lane entry, and scheduler evidence (comments, Workpad, dependencies) produce the same decisions as before.
- Candidates declined by a label-only selector produce `authorization_selector_declined` without any REST read.
- `hydrate_failed` decisions and `issue reads reached request cap` warnings disappear from a refresh of a 150-candidate board; the reconciliation stage on such a board drops well below a minute.
- Existing `TestProjectRefreshHourlyWorkload` and large-board workload tests keep passing; add the dispatch-side request assertion they currently lack (`refresh_workload_test.go:342`).

Related: #2776, #2810, #2849, #2857, #2836.

corylanou:
## Summary

ProjectV2 refresh candidates previously arrived without project fields, so dispatch re-read each issue before authorization. Include the supported field collection and timestamps in each refresh page, sharing the existing selection. Decline label-only authorization mismatches before fresh/retry hydration while retaining full authorization afterward, including nested AND/OR semantics. Declined due retries release their claim, retry, and budget refusal through the existing release helpers, preserving blocked state.

A real connector regression reproduces 150 dispatch REST reads before the fix and verifies zero afterward across two paginated refresh/dispatch cycles, with all 150 candidates reaching the production hydration hooks. Scheduler Workpad/comment/dependency evidence, custom-field decisions, and artifact timestamp decisions remain covered.

Fixes #2859

Invariants touched: [INV-3](docs/invariants.md#inv-3--mechanism-moratorium): consolidate field reads into refresh; no new mechanism, cache, configuration, or reason code.

## UI Surface Contract

- [x] N/A: no UI changes.

## Test Plan

- [x] Focused regression reproduced 150 dispatch REST reads before the fix; zero after the fix.
- [x] `go test ./internal/connector/github/... ./internal/orchestrator/...` (includes hourly and large-board workloads).
- [x] `go vet ./internal/connector/github/... ./internal/orchestrator/...`.
- [x] `make generate` (no generated changes).
- [x] `make check-fast` (assigned worker gate; no safety-critical files changed).
- [x] Rework regression reproduced retained retry/claim/budget state before the fix; blocked and unblocked retries now pass.
- [x] Rework affected package tests/vet and `make check-fast` passed (8m24s).
- [x] Exact refresh query remeasured on the Detent board: 3 GraphQL points for 100 items; hourly workload expectations updated.
- [x] Current-head CI for rework commit 6d93bf3bf: [run 35191933402](https://github.com/digitaldrywood/detent/actions/runs/35191933402) passed. PR checks are placeholders; full CI runs in the merge queue.

Production reconciliation latency requires post-deployment observation; the live instance was not modified.

Automated review was unavailable because the review bot reported its usage limit; the human review finding about retry ownership has been fixed and its thread resolved.

Rework integration repair: rebased onto main fdb7007746 and reproduced the merge-queue workload failures. The candidate field assertion now uses the fixture's lane (Todo or Human Review); measured refresh-page cost 3 is scoped to page queries, preserving synthetic PR-query cost 2. Full hourly workload, GitHub/orchestrator package tests, and vet passed; make check-fast passed in 9m54s. Current-head CI [35195248301](https://github.com/digitaldrywood/detent/actions/runs/35195248301) passed on 9ff2f0386; no actionable reviews remain.


internal/orchestrator/dispatch_planner.go
```diff
@@ -112,6 +113,16 @@ func (p dispatchPlanner) plan(
 			logDecision(dispatchPlanDecision{Issue: issue, QueuePosition: queuePosition, SkipReason: dispatchSkipMergeControlLimit})
 			continue
 		}
+		if authorization := selector.Decide(issue, labelAuthorization, p.cfg.SelectorContext); !authorization.Matched {
+			decision := dispatchPlanDecision{Issue: issue, QueuePosition: queuePosition,
+				SkipReason: dispatchSkipAuthorizationSelector, SkipDetail: authorization.Detail,
+				AuthorizationDecision: &authorization}
+			if retry, ok := dueRetries[issue.ID]; ok {
```
corylanou, line 120: Blocking: a label decline of a due retry leaks the retry and the claim. The old path ran hydrate then retryAction, which deletes state.Retry[issue.ID] and, for an authorization decline, falls through to p.releaseIssue, clearing Claimed, Retry and BudgetRefusals. The new pre-check runs before the dueRetries branch and just continues, so the retry entry and the claim stay forever, re-declined every tick; releaseMissingDueRetries only fires when the issue leaves the candidate list.

Scenario: this instance dispatches issue X labeled for it, the attempt fails retryably, the operator relabels X for another instance. This instance now holds X in Retry with a live claim indefinitely instead of releasing it, and the other instance's authorization sees a claimed issue.

Fix: on a label decline of a due retry apply the same release as retryAction does for an authorization decline (releaseClaim when Blocked, otherwise releaseIssue), and make TestDispatchLabelAuthorizationBeforeHydration retry=true assert that Retry and Claimed are cleared.

corylanou (commented):
Hand review (Codex reviewer has no quota). One blocking thread below. Non-blocking: refresh_workload_test.go:220-254 keeps the hard-coded per-page GraphQL cost of 2 "measured 2026-09-16" from the thin page without fieldValues; the page now carries a 100-node fieldValues connection per item, so re-measure and update the constant and comment so the wantPoints assertions can detect a cost regression. Also dispatch.go dispatchCandidates still hydrates without the authorization pre-check; it has no non-test callers, so fine to leave or delete.

internal/orchestrator/dispatch_planner.go
corylanou: Fixed in 6d93bf3bf. Early label declines of due retries now call releaseClaim for blocked issues and releaseIssue otherwise. The expanded regression reproduced retained Retry, Claimed, and BudgetRefusals in eight cases before the fix and now verifies cleanup plus preservation of Blocked. Focused regression, full affected package tests/vet, and make check-fast passed. Also remeasured the exact refresh query against the Detent board: data.rateLimit.cost is 3 for 100 items; updated the workload fixture and expected totals.

corylanou: Verified on 6d93bf3b: label-declined due retries release through the same releaseClaim/releaseIssue helpers retryAction uses, and the expanded regression fails on the previous head. Resolved.

PR #2878, opened 2026-09-17
fix(github): serialize refresh scans per credential

### LINKED ISSUE #2877: fix(github): concurrent board scans on one credential trip the GraphQL secondary limit in bursts; serialize scans per credential
## Evidence

One instance running seven projects on one GitHub credential (v0.114.35 then v0.114.38, 2026-09-17):

- 09:24:16Z to 09:25:08Z: `github lookup backoff scheduled trigger=github_graphql_rate_limit reason="GitHub GraphQL returned a secondary rate-limit response"` for five projects within 52 seconds (the largest board first, on `fetch github project items`, then four small boards). 
- 13:04:10Z to 13:07:14Z: the same five projects again within three minutes, `backoff_step 1` each, no restart in between (the last restart was 12:05Z).
- Between the bursts: zero backoffs for almost four hours. The bursts coincide with the projects' refresh cycles lining up (all seven `project refresh stage tracker_fetch` within the same minute).
- Cost of each burst on the busiest project: the next two ticks skip 18 to 28 candidates with `github_lookup_recovery` while the ramp reopens dispatch, so roughly 10 to 20 minutes of held dispatch per burst.
- The primary GraphQL budget was not the limit: a probe reported 4913 remaining at recovery. GitHub's secondary limit is a per-minute point and concurrency limit on the credential, so it is the sum of concurrent scans that trips it, not any one project's cost.

#2829/#2833/#2837 made the cooldown shared across the projects so one trip does not become seven; they did not change what trips it.

## Fix (consolidate, no new mechanism)

Serialize board scans per credential: a project refresh's tracker fetch and candidate hydration run one project at a time on a shared credential, in the existing refresh loop, instead of seven scans fanning out in the same minute. This removes concurrency; it adds no pacer, timer, brake, or configuration key. Keep per-project refresh cadence as is; only the GraphQL-heavy stages queue behind each other. Record `rateLimit.cost` per scan page in the existing refresh timing log line so a future burst can be attributed by points rather than inferred.

## Acceptance criteria

- Table-driven test with a fake GitHub server that counts concurrent GraphQL requests per credential: seven projects refreshing on one credential never exceed one in-flight board scan at a time; two credentials scan in parallel.
- Refresh timing lines carry the summed GraphQL points per stage.
- On the affected host after rollout: zero `github_graphql_rate_limit` backoffs over a full day of normal operation (baseline: two five-project bursts on 2026-09-17), and no increase in end-to-end refresh time for the largest board beyond the serialized wait.
- No new config key, brake, or reason code.

Related: #2829, #2833, #2837, #2859.

corylanou:
## Summary

- Queue each project's tracker fetch and candidate/status hydration behind other projects sharing its GitHub credential. Reuse the existing endpoint/credential registry, preserve refresh cadence, and release ownership before reconciliation and dispatch. Cancellation also releases queued work; different credentials remain independent.
- Exclude credential-slot queue wait from the existing frozen-tick watchdog clock while preserving elapsed real work; log `tracker_scan_wait_duration`. Cancellation ends the exclusion and completion restores ordinary cadence.
- Add response-reported GraphQL point totals to each stage in the existing refresh timing summary (`<stage>_graphql_points`). Context-scoped accounting avoids including unrelated requests on the same client.
- This consolidates concurrent scans; it introduces no new timer, backoff, recovery path, reason code, or configuration key. Existing secondary-limit handling is unchanged.

Fixes #2877

Invariants touched: none.

## UI Surface Contract

- [x] N/A: no UI change.
- [x] N/A: no visual change requiring browser verification.

## Test Plan

- [x] Seven real board fetches plus candidate hydration against a fake GitHub server: one in-flight scan per credential; two credentials progress concurrently. Bypassing scan ownership reproduces interleaved scans.
- [x] Cancellation, release after fetch failure, and per-stage point totals covered by table-driven standard-library tests.
- [x] `go test ./internal/connector/...` and `go test ./internal/orchestrator/...`.
- [x] `go vet ./internal/connector/... ./internal/orchestrator/...`.
- [x] `make check-fast` (configured worker gate; full race/coverage/nilaway gate runs in merge queue).
- [x] Long scan waits do not emit frozen-tick alerts; cancellation, fetch failure, retained stalled-work detection, and next-tick cadence covered with table-driven `testing/synctest` regressions.
- [x] Rework `make check-fast` passed in 618 seconds on `969a82829`.
- [x] Current-head PR CI [35233404379](https://github.com/digitaldrywood/detent/actions/runs/35233404379) passed on `969a82829ea0fcf71aa99bedaa03e1a90c3f5c82` (10 seconds; merge-queue deferral checks).

Human review findings addressed: queued scan time no longer triggers watchdog alerts, and GraphQL cost accounting reuses the existing response parse. The blocking thread is resolved. Automated review was unavailable because the review bot reported its usage limit; it posted no code findings.

Rollout acceptance remains to be observed on the affected host: a full day without `github_graphql_rate_limit` backoffs and no largest-board slowdown beyond serialized queue time. The worker has not modified or restarted the live instance.


internal/orchestrator/tick.go
```diff
@@ -458,6 +459,18 @@ func (o *Orchestrator) fetchTickIssues(
 	now time.Time,
 	reserve githubBudgetReserveDecision,
 ) (tickFetchedIssues, bool) {
+	if scanner, ok := o.connector.(connector.RefreshScanSerializer); ok {
+		release, err := scanner.BeginRefreshScan(ctx)
+		if err != nil {
+			recordRefreshSourceFailure(state, telemetry.RefreshSourceCandidates, err, now)
+			if o.scheduling == nil {
```
corylanou, line 472: Blocking: the time a project spends waiting for the scan slot is counted as tick time. startTick calls Advance before the wait, and the watchdog (tick_watchdog.go:96-116) flags needs_attention and logs `orchestrator tick loop frozen` when a tick exceeds two poll intervals (120 s at the configured minimum). Scenario: after a restart all seven projects start together; the seventh waits for six scans, so when the largest board's fetch takes about 60 s the later projects cross 120 s and the dashboard raises an alert for a state that is healthy and self-resolving (schedules de-phase after the first round). That is exactly the self-healing-state-as-alert pattern the operator has rejected before.

Fix: start the tick's watchdog clock after the scan slot is acquired (or exclude the measured queue wait from the tick duration the watchdog evaluates), and log the wait as part of the timing line. Add a test where a tick waits longer than two intervals for the slot and assert the watchdog does not flag it.

corylanou (commented):
Hand review (Codex reviewer has no quota). Serialization, keying by credential, release on every path, no lock across backoff or dispatch, and the fake-server test all check out. One thread to fix before merge, one nit.

Nit (non-blocking): internal/connector/github/client.go:253-260 re-unmarshals envelope.Data to read rateLimit.cost when recordRateLimitFromData (client.go:1280) already parses that field; pass the cost out of the existing parse to avoid a second full decode of large project-item pages.

internal/orchestrator/tick.go
corylanou: Addressed in 969a82829. The existing watchdog excludes only the credential-slot wait, preserving pre-wait work and post-acquisition stall detection. The wait ends on acquisition or cancellation and is logged as tracker_scan_wait_duration. A table-driven synctest regression first reproduced the false alert for a three-minute wait, then passed for success, cancellation, and fetch failure; it also verifies next-tick cadence and retained stall detection. Also addressed the review nit by recording response cost inside the existing rate-limit parse. Focused package tests/vet and make check-fast passed (618 seconds).

corylanou: Verified on 969a8282: only the scan-slot wait is excluded from the watchdog window, pre-wait work and post-acquisition stalls still count, wait logged on the timing line, tests fail with the exclusion reverted. Resolved.

PR #2849, opened 2026-09-17
fix(github): bound and resume candidate hydration

corylanou:
## Summary

- Large-board refreshes hydrate at most 25 scheduler candidates per request (13,000 connection nodes), sequentially between board pages. PR status keeps its authoritative 100-item connections and requests one PR at a time (10,401 connection nodes), avoiding premature REST fallback for larger collections.
- Completed scheduler evidence stays in the existing refresh scan across interruptions. Resume validates all comment IDs/edit timestamps and each blocker's own updatedAt before reuse. Changed or deleted blockers invalidate the dependent's evidence even when dependent and project timestamps are unchanged.
- Completed enumerations restart on the next refresh after status failures, so board lane changes remain visible. Incomplete board scans retain their cursor. Unsupported schema fields retain the existing fallback.

Fixes #2844

Invariants touched: [INV-3](https://github.com/digitaldrywood/detent/blob/100cf2c0fc141c29cffca5da8f83de7e547234a3/docs/invariants.md#inv-3--mechanism-moratorium). Consolidates hydration into the existing scan lifecycle; adds no timer, pacer, configuration key, or recovery loop.

## UI Surface Contract

- [x] N/A: no UI changes.
- [x] No persistent top-of-screen messaging added.
- [x] No visual changes requiring browser verification.

## Test Plan

- [x] Reproduced oversized hydration and operator-reported stale blocker, stale lane, and PR query-shape failures.
- [x] Connector package tests and go vet.
- [x] Initial and continuation shape ceilings, sequential hydration, interrupted batches without replay, comment edits and revision pagination.
- [x] Two consecutive status failures followed by a lane change with unchanged project updatedAt; blocker closure with unchanged dependent/project timestamps.
- [x] Changed, deleted, missing-revision, and interrupted blocker observations across bounded batches.
- [x] PRs with 51 and 100 comments, labels, reviews, contexts, and annotations remain authoritative and avoid repeated REST reads when unchanged.
- [x] make check-fast (configured worker gate; race, coverage, and nilaway run in the merge queue).
- [ ] Post-deploy operator acceptance: one hour on the largest live project with every refresh completing, zero hydrate_failed, and no CandidateHydration secondary-limit responses. No live instance mutation performed.

PR #2903, opened 2026-09-17
fix(github): bound ProjectV2 field hydration

### LINKED ISSUE #2901: fix(github): ProjectV2 enumeration page nests fieldValues per item and active_runs rescans the board outside the slot; move field reads into bounded hydration
## Evidence

One instance, seven projects on one credential, v0.114.39 (board scans serialized per credential by #2878, per-stage GraphQL points logged), 2026-09-17: four secondary-rate-limit bursts (09:24Z, 13:04Z, 13:32Z, 17:14Z). Every burst begins with the 150-candidate ProjectV2 board's refresh failing on `fetch github project items: status 403 ... You have exceeded a secondary rate limit`, then four small boards trip within 60 seconds. Per-stage points are tiny (tracker_fetch 0 to 7, reconciliation 0 to 5; the large board logged preflight 3, active_runs 35, tracker_fetch 3 before failing), the primary budget shows ~4,900 remaining, and scan waits of 1.8 s to 6.2 s prove the slot is serializing scans. Points and concurrency of scans are therefore not the trigger; GitHub's other secondary limit is server CPU time (90 s per 60 s of real time across the credential's concurrent requests), and the record fits a small number of very expensive requests.

## Cause (source-verified on `origin/main`)

1. The refresh page for a ProjectV2 board is one request enumerating up to 100 items, each with a nested `fieldValues(first: N)` connection (the selection added by #2859/#2863 so dispatch no longer re-reads issues over REST), plus Status/Priority aliases. That is a multiplicative connection: 100 items × N field values per page, evaluated server-side on every refresh whether or not any consumer needs the custom fields. Enumeration is the expensive request that fails first.
2. GraphQL that runs outside the scan slot overlaps the scan on the same credential: `active_runs` (`internal/orchestrator/tick.go:136`) calls `FetchIssueStatesByIDs` (`internal/orchestrator/reconcile.go:284`) which can populate the project-field cache with another whole-board scan (`internal/connector/github/projects_v2.go:340`); `status_drift` supplemental reads in label/issue-field modes (`internal/connector/github/lane_signal_status.go:171`); merge-completion PR hydration and review-thread reads (`internal/orchestrator/run_completion.go:1505`, `internal/connector/github/pull_requests.go:1495`); merge-queue inspection (`internal/connector/github/merge_queue.go:141`). None acquire the slot, so a second full-board scan from `active_runs` can run beside the serialized one.
3. Candidate hydration (25 aliases, `refresh_hydration.go:38`, `candidate_hydration.go:107`) and PR association/status batches (`candidate_pull_requests.go:151`, `:268`) run inside the refresh slot sequentially and are not the problem.

## Fix (consolidate, no new mechanism)

- Keep the cheap Status and Priority aliases on every board item in the enumeration page and remove the nested `fieldValues` connection from it.
- Read custom project fields only for the items that have a consumer (candidates, active states, routing states; the selection at `internal/connector/github/issue_read.go:535`), inside the existing bounded 25-alias candidate hydration, by ProjectV2 item id, so field fan-out is bounded per request and never multiplied by the page. Preserve custom values and `updatedAt` so `Fields`/`FieldUpdatedAt` still satisfy dispatch (#2859's short-circuit must keep holding: zero per-issue REST reads at dispatch).
- Make the `active_runs` project-field cache fill reuse the same bounded read instead of a whole-board scan outside the slot.
- Do not halve the page size (doubles requests, same total enumeration) and do not wrap every GraphQL call in the scan channel (deadlocks callers that already hold it).

## Acceptance criteria

- `refresh_workload_test.go` asserts the enumeration query carries no `fieldValues` connection and that custom fields are read only through bounded hydration for selected consumers (extend the assertions at ~:315 and ~:343); resume behavior and the no-per-item-fallback assertions keep passing; the #2859 orchestrator regression (150 candidates, zero REST issue GETs at dispatch) keeps passing.
- `active_runs` reconciliation never issues a whole-board scan; a test counts board-enumeration requests per tick and requires exactly one per refresh.
- On the affected host after rollout: zero `github_graphql_rate_limit` backoffs over a full day (baseline four bursts on 2026-09-17), with no increase in `hydrate_failed` decisions.
- No new configuration key, brake, or reason code.

Related: #2859, #2863, #2877, #2878, #2860.

corylanou:
## Summary

- Board enumeration retains Status/Priority and the item updatedAt scalar without nested fieldValues. Selected consumers read custom values and timestamps within existing 25-item hydration batches.
- Active-state cache misses use bounded membership and item reads instead of whole-board scans.
- Completed refreshes retain validated comment/blocker evidence. Item revisions trigger field-only reads; project timestamp changes no longer discard all scheduler evidence. PR association/status reads remain fresh each refresh.
- Removed item aliases are skipped without failing or repeatedly replaying the batch.

Fixes #2901

Invariants touched: none. Consolidates existing reads and removes blanket evidence resets; no configuration, brake, reason code, or reconciliation loop is added.

## UI Surface Contract

- [x] N/A: no UI changes.
- [x] No persistent top-of-screen messaging is added.

## Test Plan

- [x] Connector and orchestrator package suites and package go vet.
- [x] Large-board workload covers changed item fields with both stable and changing project timestamps; unchanged scheduler evidence hydrates once across four refreshes.
- [x] Bounded selected-consumer field reads, null item aliases, resume/fallback behavior, and fresh PR status reads.
- [x] Active-state reads make zero board queries; active-state reconciliation plus refresh enumerates once.
- [x] 150-candidate dispatch preserves custom values/timestamps with zero per-issue REST GETs.
- [x] make check-fast (assigned worker gate).
- [ ] After rollout: a full day with zero GraphQL rate-limit backoffs and no increased hydrate_failed decisions; local fixtures cannot establish this host-level result.


internal/connector/github/projects_v2.go
```diff
@@ -494,6 +495,10 @@ func (c *Connector) scanProjectItems(
 		// Changes during this enumeration do not prevent snapshot publication.
 		if queryDocument == thinRefreshProjectItemsQuery {
```
corylanou, line 497: Blocking: whenever ProjectV2.updatedAt differs from the previous refresh, ALL retained scheduler evidence is discarded. On a 150-candidate board with a status move nearly every tick that means re-hydrating every candidate, active and routing issue each refresh: six or more CandidateHydration requests of 25 issues with comments(first:100) bodies, blockedBy and item fieldValues, after the RefreshEvidenceRevision validation requests (refresh_hydration.go:77-82) were already spent on evidence that is then thrown away. On main, #2849's incremental retention re-hydrated only changed issues. That is a server-CPU regression in the exact dimension this PR targets, and the workload fixture hides it (constant updatedAt, no issue updatedAt in hydration responses). The inverse is also a defect: if project updatedAt does not bump on a custom-field edit, retained issues keep stale Fields indefinitely (artifact gate, claim fields).

Fix: select ProjectV2Item.updatedAt (scalar) on the enumeration page, keep comment/blocker evidence as before, and re-read fields only for items whose updatedAt changed, through the existing DetentGitHubProjectFieldHydration batch. Add a workload row where the project updatedAt changes every refresh and assert hydration requests stay bounded to the changed items.

internal/connector/github/project_field_hydration.go
```diff
@@ -0,0 +1,132 @@
+package github
+
```
corylanou, line 2: Blocking (small): a null item alias returns ErrInvalidResponse and candidate_hydration.go:121 propagates it, so one card removed from the board between enumeration and hydration fails the whole project refresh for that tick. Treat a null item like the sibling null issue alias (tolerated at :134): skip it and keep the batch.

corylanou (commented):
Hand review (Codex reviewer has no quota). The enumeration shape, bounded item-field hydration, the dispatch short-circuit and the active_runs path all check out and the new assertions fail on main. Two blocking threads; the second is small.

internal/connector/github/projects_v2.go
corylanou: Fixed in 979944424. Enumeration now selects item updatedAt; changed items use bounded field-only hydration while comment/blocker evidence remains independently validated. Successful refreshes also previously cleared evidence, so that reset is removed to preserve reuse across ticks. PR association/status reads remain fresh every refresh. The large-board regression covers changing project timestamps and field edits with unchanged project timestamps, asserting unchanged scheduler evidence hydrates once across four refreshes and only the edited item is field-hydrated. Package suites, vet, and make check-fast pass.

internal/connector/github/project_field_hydration.go
corylanou: Fixed in 979944424. Null item aliases are skipped while the rest of the batch succeeds; removed cards are not retried on subsequent pages of the same scan. Table-driven regressions cover present/removed cards and decoder behavior. Package suites and make check-fast pass.

internal/connector/github/projects_v2.go
corylanou: Verified on 97994442: evidence retained across project updatedAt changes with per-item updatedAt driving bounded field re-reads (incremental workload row fails on the previous head), null item alias tolerated. Resolved.

internal/connector/github/project_field_hydration.go
corylanou: Verified on 97994442: evidence retained across project updatedAt changes with per-item updatedAt driving bounded field re-reads (incremental workload row fails on the previous head), null item alias tolerated. Resolved.

PR #2904, opened 2026-09-17
fix(intake): scan remote default-branch revision

### LINKED ISSUE #2881: bug: stale-todos can report markers already removed from the default branch
Follow-up to #2879's separate source-root observation. The operator reported a finding for a marker removed from the default branch eleven days earlier. internal/intake/scanner.go lists tracked paths with git ls-files in source_root and reads files from os.DirFS(source_root), so it scans the current checkout rather than a defined default-branch revision.

Define and implement the intended revision for scheduled stale-todos scans without changing the live source checkout. Add a regression with divergent checkout/default-branch marker content and document the scanned revision. The marker recognition, closed dedupe, and source filters are handled in #2879; this follow-up only concerns revision selection.

No matching separate open stale-todos issue was found; #2879 explicitly identifies this as a separate observation.

```detent-agent
schema: 1
effort: low
```

UI Surface Contract: N/A.

```detent-origin
origin_kind: worker
instance_identity: CorysMacStudio.localdomain/55160
source_ref: "6026"
fingerprint: stale-todos-scans-source-root-checkout-instead-of-default-branch
```

corylanou:
## Summary

Scheduled stale-todos scans resolve and fetch the exact commit advertised by origin HEAD, then read its trees and blobs. Markers removed from the default branch no longer reappear from stale, dirty, or locally committed checkout content. The checkout, index, refs, and FETCH_HEAD remain unchanged.

Blob reads use one streaming `git cat-file --batch` process. Git terminal prompts are disabled, scans have a two-minute deadline, and failures include Git stderr. This replaces per-file subprocesses; no orchestration mechanism or configuration is added.

Fixes #2881

Invariants touched: none.

## UI Surface Contract

- [x] N/A: no UI changes.
- [x] No visual changes to primary operator screens.

## Test Plan

- [x] `go test ./internal/intake/...`
- [x] `go vet ./internal/intake/...`
- [x] `make check-fast` (configured worker gate)
- Regression coverage: divergent main/trunk and committed local-only markers, remote changes since fetch, preserved checkout/ref/index/FETCH_HEAD, absent/unreachable origin, missing remote HEAD, symlinks, oversized blobs, 0/1/2,000-file scans, context cancellation/deadline errors, and terminal-prompt override.


corylanou (commented):
Hand review (Codex reviewer has no quota). Remote-revision scanning is correct and no stale finding can be filed; three blocking threads, all small, plus a test gap.

Test gap (fix with the threads): add unreachable-remote and remote-without-HEAD cases, and a committed local-only marker case (not just a dirty worktree).

internal/intake/scanner.go
```diff
@@ -65,37 +64,23 @@ func (scannerFactory) New(name string, root string) (Scanner, error) {
...
-		}
-		if err != nil {
-			return nil, fmt.Errorf("inspect tracked stale TODO file %q: %w", path, err)
-		}
-		if !info.Mode().IsRegular() || info.Size() > maxScannedFileBytes {
-			continue
-		}
 
-		file, err := root.Open(path)
+		content, err := s.git(ctx, "cat-file", "blob", file.object)
```
corylanou: Blocking: one `git cat-file blob` subprocess per file. Measured on this repository: 2,118 scannable files, 12.4 s per scan, of which about 11.5 s is process spawns (the old code was one ls-files plus file opens). Use `git cat-file --batch` (one process fed the blob list) or `git archive` of the revision.

internal/intake/scanner.go
```diff
@@ -121,42 +106,80 @@ func (s staleTODOScanner) Scan(ctx context.Context) ([]Event, error) {
...
-		return nil, fmt.Errorf("list git-tracked files for stale TODO scan: %w; source root must be a Git worktree with git available", err)
+		files = append(files, revisionFile{path: path, object: fields[2]})
 	}
+	return files, nil
+}
 
-	paths := strings.Split(strings.TrimSuffix(string(output), "\x00"), "\x00")
-	if len(paths) == 1 && paths[0] == "" {
-		return nil, nil
+func (s staleTODOScanner) git(ctx context.Context, args ...string) ([]byte, error) {
```
corylanou: Blocking: the git helper does not set GIT_TERMINAL_PROMPT=0 and the scan runs under the manager context with no per-scan timeout (intake.go:325). On an instance running inside tmux, an https origin with a stale credential helper makes ls-remote/fetch prompt on the tty and block the scheduled loop indefinitely, stalling every scheduled intake source. Set GIT_TERMINAL_PROMPT=0 as the rest of the tree does (internal/workspace/preservation.go:165) and bound the scan with a context timeout.

internal/intake/scanner.go
```diff
@@ -121,42 +106,80 @@ func (s staleTODOScanner) Scan(ctx context.Context) ([]Event, error) {
...
+	cmd := exec.CommandContext(ctx, "git")
+	cmd.Args = append([]string{"git", "-C", s.root}, args...)
+	output, err := cmd.Output()
```
corylanou, line 245: Blocking (small): `cmd.Output()` discards stderr, so an unreachable origin logs only `git ls-remote: exit status 128` with no reason. Capture stderr into the error (trimmed) as the workspace helpers do.

internal/intake/scanner.go
corylanou: Verified on 93b76a63: single cat-file --batch (5 git processes per scan, 1.6 s on this repo), GIT_TERMINAL_PROMPT=0 with a 2-minute scan timeout, stderr in errors, new test cases present. Resolved.

internal/intake/scanner.go
corylanou: Verified on 93b76a63: single cat-file --batch (5 git processes per scan, 1.6 s on this repo), GIT_TERMINAL_PROMPT=0 with a 2-minute scan timeout, stderr in errors, new test cases present. Resolved.

internal/intake/scanner.go
corylanou: Verified on 93b76a63: single cat-file --batch (5 git processes per scan, 1.6 s on this repo), GIT_TERMINAL_PROMPT=0 with a 2-minute scan timeout, stderr in errors, new test cases present. Resolved.

PR #2910, opened 2026-09-18
fix(gate): release completed Rework waits on failed CI

corylanou:
## Summary

- Completed Rework cards whose current-head CI fails now release their gate wait through the existing `ci_not_green` routing. Previously, routing back to the same lane returned before clearing completion memory and suppressed repair dispatch indefinitely.
- Reuse completion cleanup and worker handoff without a new mechanism or redundant lane write. Pending CI and pending automated review retain their existing waits.

Fixes #2909

Invariants touched: INV-3 ([gate wait behavior](docs/invariants.md)); update the reviewed routing-function fingerprint. No new reason, configuration, or reconciliation path.

## UI Surface Contract

- [x] N/A: no UI surface changes.
- [x] No persistent top-of-screen messaging added.
- [x] No visual changes requiring browser verification.

## Test Plan

- [x] Regression reproduced `awaiting_gate` on failed CI before the fix.
- [x] Table-driven promotion-to-dispatch regression covers failed CI with and without the Rework completion marker, pending CI, and green CI with pending review.
- [x] `go test ./internal/orchestrator/...`
- [x] `go vet ./internal/orchestrator/...`
- [x] `make check-fast` (assigned worker gate)

Issue #2909, opened 2026-09-18
fix(gate): a completed card whose PR head fails CI waits at awaiting_gate forever instead of dispatching rework


corylanou:
## Evidence

A ProjectV2 project, v0.114.45/46, 2026-09-17 21:20Z to 2026-09-18 04:55Z: two Rework cards recorded `awaiting_gate` on every scheduler tick for 7.5 hours (54 and 46 repeated decisions; the repeated-decision diagnostic flagged both) while their PR heads had failing required checks (one Govulncheck failure on a head that predates the fix that landed on main, one failing Checks job). No audit ran and no worker was dispatched in that window; the cards only moved when an operator moved them to In Progress by hand.

## Cause (source, `origin/main`)

`autoPromoteActiveGatePendingIssue` (`internal/orchestrator/autopromote_tick.go:771-799`) reports a pending gate whenever `state.Completed` holds a completion record for the issue (the rework gate-wait path via `autoPromoteCompletedReworkGateWaitCurrent`, or the active-gate-eligible path), and the dispatch planner then skips the issue with `awaiting_gate`. Nothing in that path checks whether the recorded head can still pass the gate: a head whose required checks have failed cannot reach the trusted audit or the merge, so the card waits on a gate that will never open while the only thing that could fix it (a worker session) is suppressed by the same wait.

## Fix (consolidate, no new mechanism)

When the completed head's CI is failed (the same `mergeWorkerCIGreen` / CI-status evidence the gate wait already reads), the gate wait is over: clear the completion record's gate wait and let the existing Rework routing (`ci_failure_action: rework`) dispatch a session. Do not add a timer or a new reason code; reuse `ci_not_green` routing.

## Acceptance criteria

- Table-driven test: a Rework issue with a completion record awaiting the gate and a PR whose current-head CI status is failure is dispatched on the next tick with reason `ci_not_green` routing, not skipped with `awaiting_gate`; the same issue with CI pending keeps waiting; with CI green and a review pending keeps waiting.
- The repeated-decision diagnostic for `awaiting_gate` no longer accrues on failed heads.
- docs/invariants.md gate text names failed CI as ending a gate wait.

Related: #2861, #2880, #2905.

corylanou:
[...]

PR #2923, opened 2026-09-20
fix(completion): keep unfinished work out of Human Review

### LINKED ISSUE #2922: fix(completion): card promoted to Human Review while its Workpad status is in_progress and the worker reported no implementation
## Observed

On a ProjectV2 board on 2026-09-20, a card was routed to Rework at 15:35Z with an operator Workpad listing seven outstanding implementation items. A session ran 15:26:14Z to 15:36:05Z, one turn, about 1.2M tokens, and appended an honest report to the same Workpad comment:

> Rebased and lease-pushed draft PR ... No production implementation added in this pass. ... Retained generations, publication, checkpoint capture, concurrency reconciliation, alerting, and final validation remain outstanding.

The Workpad comment contains exactly one `detent-status` block and it reads `status: in_progress` with empty blockers. The pull request is unchanged: still a draft, still +166/-0 across two files, only the head SHA moved because of the rebase. Despite that, Detent recorded the attempt as terminal/success with status_message "worker completed" and moved the card Rework -> Human Review with reason `completed_active_review_transition` at 15:36:05Z.

The same card had already ended a session as success on 2026-09-09 with the same unchanged draft, so this is the second time a non-implementing session has been promoted to human review.

## Why it matters

Human Review is a request for a human to spend time. Promoting a card whose own Workpad says the work is unfinished, and whose PR is a draft the author marked do-not-merge, spends the reviewer's attention on nothing and hides the fact that the issue is not progressing. The reviewer cannot tell a finished card from an unfinished one by lane.

## Expected

The completion transition reads the Workpad status block as authoritative. A worker's terminal success does not promote a card whose latest `detent-status` says `in_progress`; that outcome routes back to the implementation lane (or records no-progress) instead. This consolidates the existing completion-signal path rather than adding a mechanism: the status block already exists and is already parsed.

## Acceptance

- A session that ends success while the Workpad status block says `in_progress` does not reach Human Review.
- A card whose linked pull request is still a draft does not reach Human Review, since a draft cannot be reviewed or merged.
- Table-driven tests for both cases, including the case where the worker appends prose after the status block.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

A successful worker session could promote an unfinished card to Human Review, while the Rework tick could bypass completion readiness and even mark a draft ready. Completion and tick routing now share the structured unfinished-Workpad check; drafts remain unchanged and ineligible for promotion.

Completion preserves genuine PR/workspace diff progress even when the Workpad remains `in_progress`. An identical-diff rebase is classified as no-progress and consumes the existing attempt allowance. Regressions verify continuation without failure evidence for changed diffs, no gate wait or promotion for unfinished work, and eventual allowance triage for repeated unchanged-diff sessions. Historical gate waits cannot supersede unfinished completion.

This consolidates existing completion evidence and removes the tick's draft-readiness mutation. No new brake, configuration, reason code, or lane writer.

Fixes #2922

Invariants touched: INV-1, INV-3, INV-5 ([same-PR documentation](docs/invariants.md)).

## UI Surface Contract

- [x] N/A: no UI changes.
- [x] No visual changes requiring browser verification.

## Test Plan

- [x] Reproduced changed-diff false no-progress and unfinished/draft Rework tick bypass before the fix.
- [x] Table-driven regressions cover both active lanes, appended prose, genuine-progress persistence and continuation without failure evidence, identical-diff allowance exhaustion, refreshed readiness, draft non-mutation, and ready controls.
- [x] `go test ./internal/orchestrator/...`
- [x] `go vet ./internal/orchestrator/...`
- [x] `make check-fast`; final invariant check after documentation correction also passed.
- [x] All four required checks passed on `3a869612ce23006031b2853fea27b312b6387286`: [CI run 35536873600](https://github.com/digitaldrywood/detent/actions/runs/35536873600). No new feedback or unresolved threads. Review bot previously reported its usage limit; local review completed.

Prior safety-critical implementation passed the 30-second safety fuzz run and race/coverage validation (81.5% overall, 91.1% implement_progress.go). Prior full `make check` reproduced baseline nilaway debt in #2924; the operator explicitly made that debt non-blocking and designated the four required CI checks as this issue's gate. This rework changes no safety-critical production file.

The existing non-blocking limitation noted in review remains: the rebase-only completion marker in memory is not restored after restart. Structured unfinished Workpads still prevent tick promotion after restart; repeated-attempt allowance history is durable.


corylanou:
Operator review: requesting changes, and the issue's acceptance criteria were amended after this session started, so please re-read #2922 before the next pass.

What this PR does: an unfinished Workpad or a draft pull request now makes `autoPromoteActiveGatePendingIssue` report a gate wait. That keeps the card out of Human Review, which was the original ask, but it does not classify the session, and that is the problem.

The concern: the card is left in a gate wait in its current lane with a Workpad that only a worker can change. In In Progress with no live worker, stranded recovery routes it to Rework and it dispatches again, which is the spend loop, the same card burning sessions with nothing delivered and nothing surfacing to a human. In Rework it risks sitting at `awaiting_gate`, which is the deadlock #2909 and PR #2910 just removed; this change reintroduces a path into that state from a different condition. Please check both paths explicitly.

What the amended issue asks for instead: classify the session. A session that ends with the Workpad still `in_progress`, or whose only change to the linked PR is a rebase that leaves the diff identical, is no-progress rather than success, and it counts against the existing no-progress and attempt-allowance brakes so the card stops on a brake and asks for a human. Blocking promotion without that classification trades a visible bad promotion for an invisible loop.

Keep the draft-PR guard, it is correct on its own: a draft cannot be reviewed or merged.

Add a test that a card which repeatedly ends with an unfinished Workpad reaches the allowance brake instead of cycling, and a test that no path introduced here can leave a card at `awaiting_gate` indefinitely.

corylanou:
Hand review (operator): requesting changes. The completion path is right this time, and the classification is genuine rather than a promotion guard: the attempt persists as no-progress and the card stops leaving the lane, so the allowance brake finally gets to act. Two things must change before this merges.

Blocking, the Workpad is not authoritative on the autopromote tick. `EvaluateAutoPromote` never reads the Workpad status, and `autopromote_tick.go:364` admits any Rework card with an open pull request through the liveRework branch regardless of the new readiness check. A Rework card carrying a structured `status: in_progress` Workpad and a green clean pull request is advanced to the pass state, Merging by default, even with the completion already recorded as no-progress. `autopromote_tick.go:153-158` also marks a Rework draft pull request ready before promoting, so the draft rule is bypassed on the same path. Apply the same unfinished-Workpad and draft checks to the tick's Rework branch, or gate liveRework on the completed-ready predicate, and add a regression that promotes a Rework card with an in_progress Workpad through the tick.

Blocking, and this one is my fault in how I worded the amended criteria, so please correct the rule rather than implement it literally. Classifying every unfinished session as no-progress is too broad: `implement_progress_test.go:441-470` shows two cases with real workspace diffs flipping from success to no-progress. Because no-progress now maps to the project failure class, genuinely progressing multi-session work would accrue project-wide failure-breaker evidence and could pause dispatch for the entire project. A false positive on a brake is worse than a missed loop. The rule should be: a session that ends with the Workpad at `in_progress` AND no change to the pull request beyond a content-preserving rebase is no-progress; a session that ends unfinished but did change the diff is progress that simply must not promote to Human Review. Keep the fingerprint comparison, which is the right instrument for exactly this.

Non-blocking: the no-progress determination survives only in memory, so a restart forgets the rebase-only case; and this PR carries two definitions of unfinished, one requiring the signal to be valid and one not. Worth reconciling while you are in here.

Nothing else. The fail-open handling for empty fingerprints, dirty workspaces, unchanged heads and merged pull requests is correct and cannot stall real work, and the tests are real rather than tautological.

corylanou:
Hand review (operator): approved. Both blocking items are genuinely fixed, verified by driving the full tick rather than reading the description.

The Rework bypass is closed. A Rework card with a structured in_progress Workpad and a green, clean, non-draft pull request now produces zero lane writes, and a Rework draft produces zero ready calls and zero promotions, while a control card with a complete Workpad still promotes to Merging, so the harness is not vacuous. The same cases fail against main exactly as described, the in_progress card advancing to Merging and the draft being marked ready. The mechanism is right too: liveRework is gated on the completed-ready predicate and re-checked after hydration, and the draft-ready mutation block was deleted rather than guarded. Your own regressions cover both.

The classification rule now requires both conditions. No-progress is applied only when the session is rebase-only, meaning same pull request, changed head, clean workspace, not merged, and an unchanged diff fingerprint. An unfinished Workpad alone no longer changes the outcome: such a session stays successful, records no failure evidence and retries, but does not reach Human Review, and it is excluded from the gate wait. The real-workspace-diff cases assert success again.

One medium finding I am NOT blocking on, but which needs a decision and I will file as a follow-up. A rebase-only session whose Workpad says complete is also classified no-progress. That penalises a legitimate conflict-clearing rebase, which is precisely the work an operator dispatches when a pull request goes dirty: it can leave the diff identical while turning a conflicting branch into a mergeable one, which is real progress by any measure that matters. Landing this now is still a clear improvement, so I would rather fix that in a follow-up than hold this.

Two small ones for the follow-up as well: the ready-marker capability now has no production caller and should go under the consolidation rule, and the rebase-only fingerprint baseline still lives only in memory, so a restart mid-session forgets it.

Issue #2925, opened 2026-09-20
test(codex): stabilize oversized-frame receive under full-suite load


corylanou:
## Observed failure

During #2919 validation on macOS, `make check-fast` failed in the existing `TestLocalTransportReceivesOversizedFrame`:

```text
--- FAIL: TestLocalTransportReceivesOversizedFrame (10.16s)
    transport_test.go:215: Receive() error = context deadline exceeded
FAIL github.com/digitaldrywood/detent/internal/codex 12.521s
```

The test launches the oversized-frame helper and uses a fixed 10-second receive deadline. Five isolated repetitions passed in 2.935 seconds total. The invariant gate, build, lint, vet, workspace tests, and changed orchestrator tests passed. No Codex transport code is changed by #2919. Root cause beyond the full-suite/isolated discrepancy is not established.

## Expected outcome

Reproduce the recorded full-suite timeout and make the test deterministic under suite load using the existing subprocess readiness/testing patterns. Preserve large-frame transport coverage and avoid production concurrency/configuration changes. Validate with the focused test and the repository gate.

```detent-agent
schema: 1
effort: low
```

Open issues were searched for `OversizedFrame` and `codex transport timeout`; no match was found.

```detent-origin
origin_kind: worker
instance_identity: CorysMacStudio.localdomain/78912
source_ref: "6113"
fingerprint: codex-oversized-frame-receive-deadline
```

corylanou:
## Detent backlog admission proposal

Evaluated at 2026-09-22T01:15:01Z. This is historical evaluation evidence; present readiness is rechecked before admission.

Proposal `admission-c6703e37aac42dd47cbb5f64` recommends moving this issue to **Todo** and meets the configured automatic-admission threshold. Detent will move the issue to that state and record this proposal as the provenance.

Criteria section: **Admission Criteria**

- **Alignment** — The recorded make check-fast timeout is an observed gate failure. The requested fix consolidates the existing test onto subprocess readiness/testing patterns, removing timing sensitivity while preserving coverage and avoiding new production mechanisms.
  - Criterion: “This includes
   flaky or timing-sensitive tests that fail CI or block a release
   cut — a red release gate is an operator-visible symptom.”
- **Readiness** — The issue records the failing test, timeout, and isolated-run comparison, and requires reproducing the full-suite failure, deterministic execution, preserved large-frame coverage, and focused plus repository-gate validation. No dependencies are declared.
[...]

PR #2926, opened 2026-09-20
fix(orchestrator): preserve completed plans across backoff

### LINKED ISSUE #2919: fix(orchestrator): preserve completed plans across coordination backoff
## Summary

A completed plan is rerun from scratch when the cross-host lane coordination write is rate limited after the plan artifacts have already been posted.

This is an orchestrator reliability bug. It must not be mitigated by lowering concurrency, changing model policy, or moving work to a weaker host configuration. The faster 24 GB Apple Silicon host must safely run the same Detent workflow and concurrency policy as the 16 GB host.

```detent-agent
schema: 1
effort: low
```

## Production evidence

Detent 0.114.49 at `bf77285a`, macOS arm64, dual-host GitHub Project coordination:

- 40 planner sessions completed successfully.
- Every work attempt then ended as `plan_transition_failed`.
- The coordination read returned `github rate limited: status 429: GitHub GraphQL rate-limit response is in backoff`.
- Detent posted 42 `## Detent Plan` comments and 40 `## Detent Plan Review` comments across three issues.
- No issue reached Plan Review.
- The failed cycle recorded 13.25 million input tokens and 112,601 output tokens.
- The project failure breaker tripped only after another six same-class failures in its one-hour window.

Project and issue identifiers are omitted because this repository is public.

## Root cause

`completePlanRunning` in `internal/orchestrator/run_completion.go` performs these operations in order:

1. `CreateComment` posts the plan artifact.
2. `updateIssueStateByID` prepares the lane write.
3. `prepareLaneWrite` calls `coordination.PublishLaneWrite`.
4. The GitHub-ref coordination store reads the shared coordination branch.
5. A 429 from that read returns `plan_transition_failed`.
6. `scheduleRetry` schedules the entire plan worker again.

The completed result is discarded as an issue attempt failure even though the failure is instance-owned GitHub infrastructure. The retry reruns the model and repeats both comments.

Current `origin/main` still has this sequence. Closed issues #1071, #2112, and #2833 cover adjacent deduplication and GitHub backoff behavior, but not replay of a completed plan after its completion fence fails.

## Required behavior

Consolidate this path into the existing durable completion-deferral and tracker-unavailable machinery. Do not add a brake, breaker, lease, park, recovery loop, reason code, configuration key, or host-specific workaround.

- Preserve the completed plan result when lane coordination or the tracker is unavailable.
- Classify the 429 as instance-owned infrastructure under INV-2, not an issue failure.
- Retry the existing completion fence after the typed GitHub retry/reset deadline.
- Make plan completion replay-safe so the existing plan and review artifacts are not recreated.
- Resume from durable state after an orchestrator restart.
- Keep INV-1 intact: the orchestrator remains the only tracker lane writer.
- Do not reduce concurrency or make the two production hosts use different workflow/model policy.

## Acceptance criteria

- [ ] A regression test reproduces a successful plan followed by a rate-limited `coordination.Store.Get`.
- [ ] Before the fix, the test demonstrates the full planner rerun and duplicate artifact behavior.
- [ ] After the fix, the runner is called once and the completed result enters the existing durable completion-deferral path.
- [ ] The work attempt is recorded as an infrastructure wait rather than `plan_transition_failed`.
- [ ] Repeated completion retries create at most one plan artifact and one automated plan-review artifact for that completion identity.
- [ ] Recovery retries only the completion fence and moves the issue to Plan Review after GitHub recovers.
- [ ] Restart recovery preserves the completed result and remains duplicate-free.
- [ ] A dual-host coordination-store test proves another host cannot cause duplicate artifacts while the first completion is waiting.
- [ ] Existing project failure breakers do not count the infrastructure wait as issue failure.
- [ ] No new operational mechanism, reason code, configuration key, or host-specific capacity reduction is introduced.
- [ ] INV-1 and INV-2 regression coverage remains green and any enforcement change updates `docs/invariants.md`.
- [ ] `make check` passes.

## UI surface contract

N/A. This changes orchestration and recovery behavior only.

corylanou:
## Summary

A successful planner previously became a failed issue attempt when publishing its Plan Review lane transition hit coordination backoff. Retrying launched the planner again and duplicated the plan and automated review.

Completed plans now checkpoint their result through existing durable completion deferral. Publication and lane-write failures keep the original attempt in an infrastructure wait, honor GitHub retry/reset deadlines, and retry completion without rerunning the planner. A completion marker reconciles uncertain comment writes; a durable receipt skips already-published artifacts. This consolidates the plan-specific failure/relaunch branches into existing recovery without adding an operational mechanism or changing concurrency/model policy.

Fixes #2919

Invariants touched: [INV-2](https://github.com/digitaldrywood/detent/blob/detent/detent-digitaldrywood_detent_2919-64fa80b68811/docs/invariants.md#inv-2--instance-owned-infrastructure-failures), documented and enforced by the added manifest regressions. INV-1 ownership remains in the existing orchestrator lane ledger.

## UI Surface Contract

- [x] N/A: orchestration and recovery only.
- [x] No visual changes to operator screens; browser verification is not applicable.

## Test Plan

- [x] Negative control against the original completion code: coordination 429 scheduled another planner call, producing four artifacts across two calls.
- [x] Fixed regression: one planner call, one plan and one review, typed reset deadline, repeated retries, no project failure strikes, and recovery to Plan Review.
- [x] SQLite close/reopen and normal durable-attempt recovery preserve the completion; a second schedule owner sharing the coordination store cannot acquire ownership during the wait.
- [x] Lost publication response, crash after comment publication, comment-read outage, tracker lane outage, and checkpoint-store failure/recovery.
- [x] `go test ./internal/orchestrator/...` and `go vet ./internal/orchestrator/...`.
- [ ] `make check-fast`: invariant checks, build, lint, vet, and workspace tests passed; the full test phase failed at unchanged `internal/codex` test `TestLocalTransportReceivesOversizedFrame` (`transport_test.go:215`, receive deadline exceeded after 10.16s). Five isolated repetitions then passed. Tracked separately in #2925; draft remains pending a green gate.
- Full `make check` race, coverage, and nilaway run in the merge queue per worker protocol; no listed safety-critical file changed.


corylanou:
Hand review (operator): approved. Verified on head fb96a4e3: build, vet, the orchestrator suite and the invariants suite all pass, so the policy fingerprint and the INV-2 doc edit agree.

The regression fixture is real rather than tautological. Applied against base, the tests fail with the production signature: runner calls 2 and artifacts 4 in both the retry and restart cases, and the crash-after-comment case shows the duplicated Detent Plan body. All pass on this head.

Restart is genuinely covered, which was my main worry. The completed result is checkpointed to the work attempt row before any remote side effect, the published flag rides through the metadata, and the restart test closes and reopens the database and recovers durable attempts before replaying. The failure path is honest too: if the durable write itself fails the code defers in memory only, and a test asserts no comment or lane write happens before the checkpoint lands. Nothing grows without bound, and it is a net removal of three bespoke outcome branches in favour of the existing deferral used by the lane ledger and permission wait, so the moratorium is satisfied.

Non-blocking, in the order I would care about them.

Publication does a comment fetch before every first-time publication, which adds a GitHub read to the exact path that is rate limited. That read is only meaningful when replaying a recovered record; on a fresh in-process completion no marker can exist. Gating it on recovery would remove one API call per plan completion, which is worth doing given what this issue is about.

The pre-publication checkpoint stamps the attempt with a deferred phase and a tracker-unavailable wait reason on every plan completion including the happy path, so the dashboard will briefly show a completion waiting on the tracker fence for plans that never waited. Short window, but it is the kind of self-healing state that should not surface as a wait.

Two theoretical ones: the comment-reader type assertion silently skips reconciliation if a connector does not implement it, which would reintroduce a duplicate in the crash window, and a local database failure is attributed to the GitHub availability signal, which matches the existing pattern rather than adding a new one.

One thing I accept rather than block on: nothing invalidates a preserved plan if the issue body or head changes during a long backoff, so a stale plan can publish. That is inherent to preserving instead of rerunning, and the alternative is the rerun this issue exists to stop.

Issue #2916, opened 2026-09-19
fix(orchestrator): selector filter records a decline decision for every fetched card in every lane


corylanou:
## Observed

After v0.114.48 (#2914, PR #2915) on an instance with an authorization selector configured, every tick records one `authorization_selector_declined` scheduler decision per fetched card regardless of lane. On the shared label-lane project this is 1823 rows per tick (all distinct identifiers) at 18:57Z, 19:14Z, 19:31Z, 19:48Z, versus 135-137 per tick before the roll (18:20Z, 18:43Z). Lane breakdown of the new rows in the first hour: Done 5388, Backlog 972, Todo 507, Blocked 356, Rework 26, Human Review 21, Cancelled 12, Plan Review 6, In Progress 4. Before the fix only Todo, Rework and In Progress cards were declined (656/26/3 over the same comparison window). scheduler_decisions volume on the instance went from ~420 rows per hour to 5478 in the first full hour. An instance without a selector is unchanged (~140 rows per hour).

The aggregate log line works as intended (`authorization_selector_declined skipped_count=1823`), but the per-card rows now cover the whole board fetch, including closed and Done cards that no write path would ever touch.

## Cause

`internal/orchestrator/authorization_tick.go` filters the full fetched set (candidates plus retained inputs such as board issues and lane-signal candidates) at `internal/orchestrator/tick.go:142` and records a decision for each excluded identity. The dispatch planner previously only evaluated cards in dispatchable lanes, so declines were bounded by those lanes.

## Expected

Keep the single filter, but record per-card decline decisions only for cards the planner would have evaluated (dispatchable lanes), matching the pre-#2914 row count; excluded cards in other lanes are covered by the aggregate skip count alone. No new mechanism, no configuration key, no retention change.

## Acceptance

- With a selector configured, a tick over a board with N Done/Backlog/closed unowned cards and M unowned dispatch-lane cards records M decline decisions, not N+M.
- The aggregate skipped_count still reports N+M.
- Table-driven test covering mixed lanes.

Depends on #2914

```detent-agent
schema: 1
effort: low
```

corylanou:
## Codex Workpad

Implemented and validated: https://github.com/digitaldrywood/detent/pull/2918 (ready for review).
Head: 5a8256f5f7e22e030d4b8c24afd53df1ba085c63.

[...]

Issue #2914, opened 2026-09-19
fix(orchestrator): recovery and reconciliation paths write lanes on issues the authorization selector excludes


corylanou:
## Observed

A label-lane project is shared by two Detent instances, each configured with an authorization selector on a per-instance label (instance A requires `detent:a`, instance B requires `detent:b`). On 2026-09-19 instance A's scheduler declined every card that lacks its label (`authorization_selector_declined`, 179 cards per tick), yet the same instance wrote lane changes and comments on a card that carries only instance B's label:

- 16:26Z: instance A recorded In Progress -> Todo with reason `stranded_active_recovery` on that card while instance B's worker session was live on it (B's Workpad was updated at 16:37Z and posted a human question).
- 16:48Z: instance A moved the card Todo -> Human Review with reason `disabled` (`reconciliation: stale_todo_pr`, linked PR by instance B), logged `stale_todo_pr_reconciled` at 16:55:40Z, and posted "Reconciled this issue from Todo to Human Review because it already has a linked PR" under instance A's operator identity. Instance B then posted the same reconciliation comment twice (16:56Z, 16:59Z), so the card carries three identical comments.
- Instance A also evaluates `blocked recovery decision` and `dependency auto-unblock decision` for cards it does not own.

Same shape on two more cards the same day: a human moved a card to In Progress, instance A bounced it to Todo with `stranded_active_recovery` within 25 minutes because it had no local attempt (04:18Z/04:42Z and 15:59Z/16:26Z).

## Cause

The selector is enforced only in the dispatch planner (`internal/orchestrator/dispatch_planner.go:118`). The candidate fetch passes an `IssueFilterHint` (`internal/orchestrator/tick.go:643-665`) but the fetched candidate set still contains every card in the lanes (the 179 declines prove it). `recoverStrandedActiveIssues` (`internal/orchestrator/tick.go:316`, `internal/orchestrator/stranded_active.go:64`), the stale Todo PR reconciliation (`internal/orchestrator/autopromote_tick.go`, `applyStaleTodoPullRequestDecision`), blocked recovery and dependency auto-unblock all iterate that set and call the tracker writer without consulting the selector, so an instance that is not authorized to work a card still rewrites its lane and comments on it.

## Expected

[...]

PR #2927, opened 2026-09-20
fix(github): filter ProjectV2 refresh before enrichment

### LINKED ISSUE #2917: Apply instance filters before ProjectV2 refresh enrichment
## Observed behavior

On Detent v0.114.47, the MacBook Air instance's ProjectV2 refresh failed before delivering its expected Todo candidates to local scheduling.

The Parable project contained 484 open issue cards: 248 Backlog, 132 Todo, 89 Blocked, 6 Human Review, 6 Rework, 1 In Progress, 1 Done, and 1 Cancelled.

The instance selects issues labeled `detent:macbook-air-1`. Eight open issues carried that label. Its five expected Todo matches were:

- getparable/parable#3531
- getparable/parable#3485
- getparable/parable#3481
- getparable/parable#3480
- getparable/parable#1604

One refresh recorded 1,650 REST calls:

- 518 issue reads
- 468 dependency reads
- 335 issue-comment reads
- 323 PR lookups
- 6 other requests

The PR lookups checked for PR absence. They do not represent 323 existing PRs.

Telemetry also reported 1,406 HTTP 304 responses and 150 GitHub-billable requests. These are separate measurements, not a complete partition of the 1,650 calls. The observed comment-call count also does not establish the scope of the local cap counter.

The refresh failed with Detent's local request-cap error:

```text
fetch dispatch Workpad comments failed: fetch github issue comments: github rest fanout deferred: endpoint family issue comments reached request cap 500
```

This failure was a local cap rejection, not a GitHub quota rejection.

## Cause

At main commit `43f09e72205ce4d61f208efa6810d3c600ded33d`, ProjectV2 `FetchRefreshIssues(ctx, candidateStates, observedStates, hint)` calls:

```text
fetchProjectRefreshIssues(ctx, candidateStates, observedStates, hint.SchedulerStates)
```

It drops `Authors`, `Assignees`, `LabelInclude`, and `LabelExclude`.

The combined refresh therefore performs per-issue enrichment across a broader set than the instance can select. Later scheduler filtering cannot recover the requests already spent.

v0.114.48 retains this behavior. Its #2914 fix restricts unauthorized recovery and reconciliation writes after fetch; it does not remove this connector fetch amplification.

## Requested change

Carry the complete instance eligibility predicates into the existing ProjectV2 combined refresh and evaluate them before per-issue enrichment for blockers, dependencies, comments, and PR evidence.

Use the existing ProjectV2 GraphQL inventory and batched hydration paths. Preserve the current predicate combination semantics. Fetch all metadata required to evaluate a predicate, including nested label or assignee pages where necessary. Missing or incomplete metadata must not silently broaden eligibility.

An issue's membership in an observed lane alone must not exempt it from filtering. If existing tracked work requires observation outside the selector, identify the existing condition and retain only that necessary observation. Likewise, retain evidence about an excluded dependency when an eligible issue requires it. Neither case makes the excluded issue a dispatch candidate or grants permission to mutate it.

Apply the same filtering to existing fallback paths. Reuse existing local SQLite state where useful, without treating a locally cached issue or an unchanged issue timestamp as proof that comments, native dependencies, or PR evidence remain current.

Keep this change within the current refresh path. Preserve existing persistence and failure semantics, including any existing guarantees that failed refreshes retain last-good records and do not dispatch from incomplete evidence. Do not introduce partial-success scheduling as part of this fix.

## Storage and scope

Current local SQLite state is available for reuse. A durable normalized tracker snapshot or revision table is a separate design option, not a prerequisite or deliverable here.

If the implementation reveals a need to change persistence or partial-refresh behavior, file that product change separately with `detent:todo`. Any later snapshot proposal must identify the repeated work it replaces within the existing refresh path.

The mechanism moratorium applies. Do not add brakes, breakers, leases, recovery paths, reason codes, reconciliation loops, configuration keys, CLI commands, or dashboards. Do not raise request caps to make the fixture pass.

## Acceptance criteria

- [ ] A regression fixture contains 484 open project cards with the observed lane distribution, eight issues matching `detent:macbook-air-1`, and exactly the five named Todo matches. Give those five complete evidence that permits candidate delivery.
- [ ] Exercise the production combined refresh path through candidate delivery and assert the exact five repository-qualified candidate IDs. This test fails against the pre-fix implementation.
- [ ] Record REST requests by endpoint family and target issue. With no required observed work or cross-issue dependencies, excluded issues receive zero per-issue enrichment requests.
- [ ] Compare that fixture with a smaller board containing the same eight selected issues and identical evidence. Adding excluded cards produces zero additional per-issue REST enrichment calls. Assert fixed numeric enrichment counts or bounds derived from the fixture, and keep every capped family below its existing limit.
- [ ] Cover author, assignee, label-inclusion, and label-exclusion predicates, including existing combination semantics. Excluded issues are filtered before enrichment.
- [ ] Cover project-card pagination, incomplete selector metadata, and nested metadata pagination where applicable. Exercise existing fallback paths and verify that they preserve selection.
- [ ] Cover existing required observation and dependency-evidence exceptions, if any. Name the exact issue IDs and assert their request counts. Unrelated observed-lane cards remain unenriched, and excluded issues do not become dispatch candidates or gain write authorization.
- [ ] Change comments, native dependencies, and PR state between refreshes without changing the issue's own update timestamp. Verify that the existing refresh cadence still brings each change into the evidence used by scheduling.
- [ ] Cover HTTP 304 responses and distinguish actual request counts, local cap accounting, and GitHub-billable usage.
- [ ] Inject an enrichment failure after partial progress. Verify that this change preserves existing refresh-failure and persistence guarantees and cannot newly enable dispatch from incomplete evidence.
- [ ] Report before-and-after request measurements with the measurement window and counter scope stated. Confirm that the fix uses the existing refresh path and adds none of the prohibited mechanisms.

## Related issues

- #2761: Batch board and native dependency hydration.
- #2776: Reuse batched hydration in normal project refresh.
- #2859: Fill ProjectV2 fields and decline labels before dispatch hydration.
- #2914: Authorize recovery and reconciliation candidates, released in v0.114.48.

These issues are closed. No open duplicate was found during issue search.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- Apply instance author, assignee, required-label, and excluded-label predicates before ProjectV2 scheduler enrichment, including existing REST fallbacks. Complete nested selector metadata first and retain excluded inventory only for dependency lane evidence and diagnostics.
- Remove completed-scan scheduler-evidence reuse so independent comment, native dependency, and PR changes are observed without relying on issue timestamps. Interrupted scans retain their existing resume validation; persistence and refresh failure behavior are unchanged.
- Reproduce the 484-card incident distribution and deliver exactly the five named candidates through the production dispatch planner. In one cold fallback refresh, original code makes 265 REST requests (132 comments + 132 dependencies + one repository PR list); fixed code makes 11 (5 + 5 + 1). The eight-card board has identical fixed counts. Complete batched evidence makes one REST PR-list request. These fixture measurements are separate from the reported incident's 1,650 calls.

Fixes #2917

Invariants touched: INV-1 — [same-PR authorization boundary documentation](https://github.com/digitaldrywood/detent/blob/5df46940a5756a8d24c0fb01f66f74966d8716c7/docs/invariants.md#inv-1--lane-ownership). Excluded cards receive no scheduler enrichment or blank-status repair authorization. Existing direct-ID observation of tracked work is unchanged; observed-lane membership creates no exception. No new mechanisms, configuration, storage schema, CLI, dashboard, or request caps are introduced; completed evidence reuse is removed rather than adding a freshness guard.

## UI Surface Contract

- [x] N/A: no UI surface changes.

## Test Plan

- [x] Regression fails with original production files restored: 132 candidates instead of the exact five repository-qualified identifiers.
- [x] Connector package tests and vet; focused orchestrator dispatch, authorization, and last-good board retention regressions.
- [x] Author/assignee alternatives, AND combination, all inclusion labels, any exclusion label, nested labels/assignees, incomplete metadata, card pagination, both scheduler fallbacks, and excluded blank-status repair.
- [x] Selected observed routing #9002 and #9003: seven comments + seven dependencies + two PR lists = 16 calls on either board; unrelated observed cards remain unenriched. Excluded dependency fixture/fresh#2 has zero independent enrichment calls and is never a candidate.
- [x] Comments, native dependencies, and PR evidence change between refreshes with unchanged issue timestamps; partial-progress failure publishes no incomplete candidate/status/diagnostic snapshot.
- [x] Two ETag fallback refreshes: 22 actual requests, 11 HTTP 304, 11 billable. Second per-refresh local-cap consumption: 11 quarter-request units (2.75 request equivalents); existing cap 500 unchanged.
- [x] `make check-fast` (assigned worker gate; race, coverage, and nilaway remain merge-queue checks).

Current-head PR checks: [CI run 35532640953](https://github.com/digitaldrywood/detent/actions/runs/35532640953) is green. These PR checks are acknowledgments; substantive CI runs in the merge queue. The Codex review bot reported exhausted review quota and produced no code review; there are no actionable review threads. Human approval remains required before merge.


corylanou:
Hand review (operator): approved. Verified on head 5df46940: build, vet and the connector plus orchestrator suites pass, and the new tests fail against base bf77285a (the 484-card variants return 132 candidates instead of 5), so the fixture is a real regression test rather than a tautology.

Checked the exclusion risks specifically. The connector predicate uses only authors, assignees and label include/exclude, which is a strict relaxation of the full selector, so nothing the connector drops would have survived `selector.Match` anyway, and post-enrichment field or assignee predicates cannot be defeated. Owned cards in Merging and Human Review still match. Dependency blockers do not freeze: blocker state still resolves from the full board and arrives inside the owned card's own batched hydration. It does not collide with #2914 either, since lane-signal candidates stay unfiltered here and are re-filtered with the full selector at the tick boundary, which is what preserves the decline diagnostics #2916 just bounded.

Two non-blocking notes for the record.

First, and worth a line in the PR description and the release note: behaviour on a board with NO selector configured is not unchanged. The completed-scan retention is discarded unconditionally, so every refresh re-hydrates all scheduler-lane cards; the test deltas show it, GraphQL points 241 to 328 and hydrations 1 to 3. That is GraphQL cost rather than the REST cap that caused the incident, and the issue's freshness criterion sanctions it, but instances without a selector will pay it every tick and should not discover that from a graph.

Second, low: a connection reporting totalCount greater than the node count with hasNextPage false fails the whole refresh with zero candidates rather than excluding the one anomalous card. Fail-closed is the right default here, so this is only a note.

PR #2928, opened 2026-09-20
fix(validation): resolve baseline NilAway findings

### LINKED ISSUE #2924: fix(validation): resolve existing nilaway findings blocking make check
## Observed

The full `make check` gate for #2922 fails in `nilaway-audit` with 15 primary diagnostics. Running `make nilaway-audit` against a clean archive of the prior commit `ab17b9a99129c6a2d40fc3d61516381925e30239` reproduces the identical set; the completion-classification rework adds no findings.

Affected sources include `internal/intake/scanner.go:250`, GitHub connector flows from `internal/connector/github/issue_read.go` through `pull_requests.go` and `statuslabel.go`, and `internal/runner/agent_override.go:216` via `provider_capacity.go`. Several findings concern nil slices; establish actual defects versus analyzer false positives before fixing.

## Acceptance

- Reproduce with the repository-pinned nilaway version `v0.0.0-20260612163715-2d8907f431ca`.
- Resolve the baseline findings without disabling the gate or broadening issue #2922.
- `make nilaway-audit` passes and the full validation gate can advance.

Origin: machine-discovered during #2922 validation. Search of open nilaway issues found no matching report.

```detent-agent
schema: 1
effort: low
```

```detent-origin
origin_kind: worker
instance_identity: CorysMacStudio.localdomain/78912
source_ref: "6114"
fingerprint: nilaway-baseline-nil-slice-and-provider-override-findings
```

corylanou:
## Summary

- Resolve the 15 baseline findings reproduced with pinned NilAway `v0.0.0-20260612163715-2d8907f431ca`, without exclusions or suppressions. Connector findings were false positives from indexes built alongside nil-initialized slices; initialize those slices explicitly. The override finding followed the address of a value struct; return the updated value instead.
- Require the scanner command name as an argument, eliminating error-path indexing into variadic arguments. Existing callers already supply it.
- Correct the fast-gate message: it excludes race, coverage, and NilAway; `make check` provides full local validation. No current GitHub workflow runs NilAway.

Fixes #2924

Invariants touched: none. No mechanisms, lane transitions, or safety-critical orchestrator files changed.

## UI Surface Contract

- [x] N/A: no UI changes.

## Test Plan

- [x] Reproduced all 15 primary diagnostics with the repository-pinned NilAway version.
- [x] `make nilaway-audit` passes after fixes.
- [x] Tests and `go vet` pass for intake, GitHub connector, and runner packages, including a table-driven nil/empty refresh-batch regression and existing override fallback regressions.
- [x] `make check-fast` passes (configured worker gate, including invariants, lint, build, vet, and tests).
- Full `make check` is deferred under the worker validation contract; no safety-critical orchestrator files changed.


corylanou:
Hand review (operator): approved. Verified on head ae9de9e3 in a clean clone: build and vet clean, the connector, intake and runner suites pass, and `make nilaway-audit` now reports zero diagnostics where the baseline had fifteen.

Checked the two changes that are not purely mechanical. The scanner's git helper takes the operation as its own parameter instead of indexing args[0], and every call site already passes a subcommand first, so behaviour is identical and a zero-argument call is now a compile error rather than a panic. The agent override helper moved from mutating through a pointer to taking a value and returning it; both call sites were updated to reassign, the role map takes addresses into the local copy so the mutations land on the returned value, and only string fields are cleared, so the shallow copy is safe. A missed call site would have failed the build, and it builds.

The nil slices that became empty slices are internal to their functions and nothing compares them against nil or marshals them, so there is no nil-versus-empty behaviour change. The added test for empty refresh batches asserts no HTTP request is made, which is the right shape.

Also correct to fix the Makefile line in the same change. The old text claimed race, coverage and NilAway run in the merge queue, which is not true of any workflow under .github, and that claim is what led an earlier session to treat these pre-existing findings as blocking.

PR #2932, opened 2026-09-21
fix(github): filter ProjectV2 refresh scans by instance selector

### LINKED ISSUE #2930: fix(github): ProjectV2 board scan costs 45 GraphQL points per refresh and stalls every project on the host
Follow-up to #2917, which reduced REST enrichment. This is the GraphQL side and it is measured after #2917 shipped, not before.

## Measured

On one host running seven projects against a single GitHub credential, refresh cost in GraphQL points (tracker fetch plus status drift), sampled after v0.114.51 was live:

| project | refreshes | avg points | max |
| --- | --- | --- | --- |
| large ProjectV2 board | 2 | 45.0 | 45 |
| next highest | 4 | 1.2 | 2 |
| next | 4 | 1.2 | 2 |
| next | 4 | 0.5 | 1 |
| remaining two | 7 | 0.0 | 0 |

The large board's pre-#2917 baseline over 23 refreshes was 37.4 average with a max of 46, so #2917 did not change this number, which is expected: it filtered enrichment requests, not the board scan.

Every tick on that board also records 137 `authorization_selector_declined` decisions, against roughly nine cards the instance actually owns, so the scan enumerates the whole board to find a small owned subset.

## Impact observed

Between 21:54Z and 23:08Z on 2026-09-20 the host logged a GitHub GraphQL secondary rate-limit response and every one of the seven projects sat in a `github_lookup` dispatch recovery at `limit: 0`, with the large board additionally in a `pull_request_hydration` recovery with reason `secondary_throttled`. That is 22 consecutive ticks with zero dispatch across the host, self-healing only in the sense that each backoff re-arms roughly one tick later. Two ready cards on the large board made no progress for over an hour.

## What this issue does NOT claim

It does not claim the secondary rate limit and the REST amplification in #2917 share a cause; they are different quotas and the link is unproven. It does not equate request counts with billable requests. The only causal claim is the recorded one: the recovery's own reason field says a GraphQL secondary rate-limit response was returned.

## Expected

The per-refresh GraphQL cost of a board should scale with the cards an instance can act on, not with total board size. Apply the authorization selector inside the board scan, in the same spirit as #2917 applying it before enrichment, so a board of several hundred cards with nine owned ones does not cost forty-five points every refresh. No new mechanism, no configuration key: the selector and the filter hint already exist.

## Acceptance

- Refresh points for a large board with a narrow selector fall to the same order as the other projects on the host.
- Cards the instance owns are still delivered, including blocked cards whose blockers live outside the selector.
- A test asserts points scale with owned cards rather than board size.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- ProjectV2 refresh now applies representable label and assignee selector hints inside `items(query: ...)`, so unrelated cards no longer drive board pagination. The local selector remains authoritative; author predicates and ambiguous search values stay local.
- Preserve external blocker lanes using targeted reads through the existing project-item reader, including closed and human-owned blockers. Retryable blocker lookups retain unresolved evidence without withholding unrelated cards. Interrupted scans restart when the filter changes.
- Regression fixture: 1,991 cards with nine owned falls from 20 pages/40 estimated points to one page/two points; 109 owned requires two pages/four points. A read-only probe of the affected Parable board returned 42 matching labeled cards in one page at GitHub-reported cost 2. The live process was not changed; this measures scan cost, not a deployed full-refresh result or the cause of secondary throttling.

Fixes #2930

Invariants touched: [INV-1](https://github.com/digitaldrywood/detent/blob/37c9c8cca3c212f30c3e7e6770701e4c5891cc87/docs/invariants.md#inv-1--lane-ownership), updated in this PR with filtered enumeration and external dependency evidence. This consolidates selector filtering into the existing scan and reuses the dependency reader; no new operational mechanism or configuration is introduced.

## UI Surface Contract

- [x] N/A: no UI changes.
- [x] No visual changes requiring browser verification.

## Test Plan

- [x] Reproduced the unfiltered scan amplification before implementing the fix.
- [x] `go test ./internal/connector/github/...`
- [x] `go vet ./internal/connector/github/...`
- [x] Table-driven regressions for cost, filter semantics, pagination, external blockers, human prerequisites, retryable failures, and changed-filter resume behavior.
- [x] `make check-fast` (assigned worker gate; race, coverage, and nilaway belong to the merge queue).


corylanou:
Hand review (operator): approved. Build, vet, the connector and orchestrator suites and the invariants suite all pass, and the filter semantics were confirmed against the live boards rather than assumed: repeated label qualifiers AND, comma-joined values OR, negated labels exclude correctly, and an unknown label yields zero rather than an error.

The exclusion risk is clear, which was my main concern. The hint is built from the top-level selector only and the scan predicate projects a subset of that, dropping any value it cannot safely quote, so it always widens rather than narrows and cannot exclude a card the full selector would keep. Authors stay local, correctly, since project filters have no author qualifier. Cards in Merging and Human Review and those with in-flight attempts are label and assignee stable, so they survive.

Dependency blockers, the case that would silently freeze an owned card, are handled: off-board blockers get their project item read explicitly, closed blockers already map to a terminal state upstream, and a failed lookup leaves the state empty, which the blocked-recovery path treats as still blocking. That fails safe rather than dispatching on a blocker whose state is unknown.

The measurement is real and asserted through the full fetch path rather than on the filter function: a 1991-item board with nine owned cards costs one page and two points, the same as a nine-item board, against twenty pages and forty points unfiltered. That is the number this issue was filed on.

Three non-blocking notes. The scan now aborts as truncated if a selector label or assignee changes mid-enumeration, because the first-page total is filter-sensitive; it self-heals on the next tick but it is a new failure trigger and deserves a line in the description. Blocker-state resolution now has a third path alongside the existing two, and the moratorium asks a mechanism-touching change to consolidate or say why it cannot, so please say which. And the fallback state is cached under the blocker id seeded from the first referring reference, which is harmless today because both derive from the same blocker, but is fragile if that stops being true.

PR #2942, opened 2026-09-21
fix(make): serialize gates and allow lint headroom

### LINKED ISSUE #2937: fix(make): make the gate deterministic under concurrent workers; serialize check-fast, raise the lint timeout
## Evidence (detent project, 2026-09-15..21, Codex rollouts)

Sessions spent 29.7 hours of 78.5 hours of wall-clock inside the gate, and 18.6 of those hours were second-and-later runs. Reading 62 repeat runs across the 12 worst sessions, the dominant cause is not the agent and not the code:

- 43.5% environment, not a real failure: Go test-cache input scanning stalls (#2735), missing build-cache entries after the host cache trim, `golangci-lint` exceeding its own 5m timeout at 394s while reporting zero findings, and tests with 1-second deadlines flaking under host load.
- 22.6% a genuine failure the agent then fixed. That is the gate working.
- 14.5% a lint or invariant sub-gate failing on its own.
- 8.1% revalidating after a rebase.
- 3.2% the gate passed and the agent ran it again anyway. "Ran after every small edit" is zero.

Contributing mechanism: `make check` serializes across worktrees through `tools/checklock` (Makefile:191) but `check-fast` does not (Makefile:197), and concurrent detent sessions peaked at 9 this week. So the target agents are told to use is exactly the one with no concurrency control, on a host running up to 9 of them at once.

Sub-step timing: median gate 651s, p90 1074s. About 418s is `go test` (internal/workspace 122s median; internal/connector/github and internal/cli own the tail) and about 233s is invariants, migrations, generated, build, lint and vet, with lint the largest non-test step.

## Change

1. Extend the `checklock` serialization to `check-fast` (Makefile:197), or pin `GOMAXPROCS` per worker so concurrent gates cannot starve each other. Whichever is chosen, the two targets must have the same concurrency behavior.
2. Raise `golangci-lint --timeout` (Makefile:166) above its observed p90; 5m is below what it actually takes under load, so it fails for no reason and the agent reruns.
3. Give the 1-second-deadline tests that flake under load a deadline proportional to available parallelism, or mark them to skip when the host is saturated.

## Expected effect

Removes roughly 12 of the 18.6 hours per week that currently go to repeat gate runs, and with them the tokens spent waiting on those runs. Measured by the same rollout query: repeat gate runs per session and wall-clock inside the gate.

## Acceptance

- Repeat gate runs classified as environment drop below 10% of gate runs.
- `golangci-lint` timeout failures with zero findings reach zero.
- Gate wall-clock share of session time falls from 37.8%.

```detent-agent
schema: 1
effort: low
```

corylanou:
## Summary

- Route `check` and `check-fast` through the same existing FIFO validation lock across worktrees, preserving generated-input preflight and rechecking after admission.
- Raise lint timeout from 5 to 15 minutes to allow headroom above the reported 394-second zero-finding run.
- Give workspace shell-hook completion tests a 10-second minimum budget scaled by test parallelism relative to GOMAXPROCS. Keep intentional timeout assertions unchanged and skip no tests.

Fixes #2937

Invariants touched: none. This consolidates gate admission through the existing mechanism and changes no orchestrator behavior.

## UI Surface Contract

- [x] N/A: no UI changes.

## Test Plan

- [x] Extended Makefile regression fails against original check-fast: admission is bypassed. Covers both gates, shared lock path, stale inputs, changes during admission, and invariant failures.
- [x] Checklock tests, changed workspace tests, and root-package tests pass. The full workspace suite passed in the first full gate.
- [x] `go vet . ./tools/checklock ./internal/workspace`
- [ ] `make check-fast`: blocked by unrelated intermittent `TestGitFixtureEnvironmentIsolation/worktree_and_index`: `copy source repo seed: open .git/objects/maintenance.lock: no such file or directory`. All other steps passed in the latest run; race/coverage/NilAway are merge-queue validation.

Fleet acceptance rates and wall-clock improvement require post-deployment observation; this PR does not claim those metrics have already improved.

Validation history: the initial focused workspace suite hit a transient Git fixture `maintenance.lock` copy race. Filing a separate follow-up via file_machine_issue was rejected by the instance repository REST request cap. The first full gate passed workspace but found lint-contract assertions still expecting 5m; those assertions and contributor docs are now updated and root tests/vet pass.


corylanou:
Hand review (operator): approved. Build and vet clean; the workspace and checklock suites pass. The full-suite failures in the CLI token-audit tests reproduce identically on the base revision in a clean clone, so they are pre-existing environment artifacts rather than anything this change caused.

The premise holds and, importantly, this is not a new mechanism. The lock tool already exists on main with its queue, activity tracking and recovery; this change only touches its README and tests. What it actually does is merge the two gate entry points into one rule that runs the existing lock, so the fast gate stops being the one path that bypassed admission. That is consolidation, which is what the moratorium asks for, rather than a new brake.

The wait is bounded rather than indefinite, which was my main concern: fifteen minutes of no-progress budget, reset by owner handoff, output activity or queue advancement, under a four hour ceiling, and on expiry it exits with an acquire error so a starved session fails visibly instead of hanging.

Blast radius is local. No workflow invokes the combined gate targets; CI calls the sub-targets directly. The one thing that does reach CI is the lint timeout going from five to fifteen minutes, which is justified by the observed 394 second zero-finding run cited on the issue, though it means a genuinely hung lint now burns fifteen minutes before failing.

Tests are real rather than string-greps: they run the gate targets in a temporary directory with stubbed tooling and assert an ordered execution trace and the lock path, and actual mutual exclusion is already covered by the existing serialization and overtaking tests.

One thing to watch after this rolls, not a blocker. With up to nine concurrent sessions and a median gate around 651 seconds, a deep queue can approach the four hour ceiling, at which point a wait becomes a failure. Keep an eye on the wait-failed events in the validation event log; if they appear, the answer is fewer concurrent gates rather than a longer ceiling.

